PR #110, opened 2026-04-12
feat: add sidebar toggle for collapse to icon-only mode

Trinathswain45:
# Summary
Adds a toggle button in the sidebar header to collapse the sidebar to an icon‑only view, reducing width to save space on smaller screens. Navigation labels and user info are visually hidden while icons remain accessible, and the main content padding adjusts accordingly.


Trinathswain45:
Hi @SdSarthak , Pls review and let me know if any changes required.

PR #164, opened 2026-05-10
Add ignoreDeprecations to silence the baseUrl deprecation

Sellamuthu2007:
## Summary

Added "ignoreDeprecations": "6.0" to [tsconfig.json](vscode-file://vscode-app/c:/Users/R.SELLAMUTHU/AppData/Local/Programs/Microsoft%20VS%20Code/8b640eef5a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to silence the TypeScript 7.0 baseUrl deprecation warning. This allows the project to compile without errors while maintaining current path aliasing.

## Type of Change
- [ ] Infra / CI
# Checklist
 - [ ] I have read CONTRIBUTING.md
 - [ ] pytest backend/tests/ passes locally (not applicable)
 - [ ] I have not committed .env or any secrets

### DIFF
--- a/backend/tests/test_bulk_import.py
+++ b/backend/tests/test_bulk_import.py
@@ -3,8 +3,7 @@
 import pytest
 from unittest.mock import MagicMock, patch
 from fastapi.testclient import TestClient
-from io import StringIO
-
+from io import BytesIO
 
 @pytest.fixture
 def client():
@@ -42,8 +41,8 @@ Fraud Detector,Flags anomalous transactions,Risk Assessment,Finance,2.1"""
 
         response = test_client.post(
             "/api/v1/ai-systems/import",
-            files={"file": ("test.csv", StringIO(csv_content), "text/csv")}
-        )
+                files={"file": ("test.csv", BytesIO(csv_content.encode('utf-8')), "text/csv")}        
+                )
 
         assert response.status_code == 200
         data = response.json()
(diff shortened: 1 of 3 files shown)


SdSarthak (changes requested):
Thanks for the contribution, but this approach has two issues:

1. **`ignoreDeprecations` suppresses the warning rather than fixing it** — adding `"ignoreDeprecations": "6.0"` tells TypeScript to silently accept deprecated options instead of removing the deprecated option. The correct fix for the `baseUrl` deprecation warning is to migrate path aliases to the Vite config (using `vite-tsconfig-paths` or a manual `resolve.alias` in `vite.config.ts`) and remove `baseUrl` + `paths` from tsconfig entirely. The project will eventually upgrade to TypeScript 7 and this will break.

2. **Changing `baseUrl` from `"."` to `"src"` is a separate change** with its own implications — it alters how non-relative imports are resolved. Please do not bundle this into the same PR as a warning-suppression fix. If you want to propose this change, open a separate issue so the impact can be evaluated.

If you'd like to properly fix the TypeScript path alias setup, the path forward is:
- Install `vite-tsconfig-paths` (or configure `resolve.alias` in vite.config.ts)
- Remove `baseUrl` and `paths` from tsconfig.json
- Revert this change entirely

Happy to point you to resources if that would help.

Sellamuthu2007:
@SdSarthak

Thank you for pointing to the complete fix.

Happy to collaborate.

SdSarthak:
Closing due to merge conflicts. Note: `ignoreDeprecations` should be `"5.0"` for TypeScript 5.x, not "6.0".

PR #243, opened 2026-05-12
fix: add required validation for registration fields

Khyati-8180:
**Changes Made**

-  Added required validation for Full Name field
- Added required validation for Company Name field

**Problem**

Previously, the registration form allowed users to submit without entering Full Name and Company Name, causing inconsistent validation behavior.

**Result**

The form now properly validates all required fields before submission.

### DIFF
--- a/frontend/src/pages/Register.tsx
+++ b/frontend/src/pages/Register.tsx
@@ -85,6 +85,7 @@ export default function Register() {
             <input
               id="full_name"
               type="text"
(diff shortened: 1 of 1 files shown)


SdSarthak (approved):
Adding `required` to the `full_name` and `company_name` inputs is correct — it gives immediate browser-level validation feedback before the form is submitted. Simple, correct fix. Approving.

PR #245, opened 2026-05-12
Add bulk classification endpoint with partial failure handling

### LINKED ISSUE #51: Add bulk classify endpoint — POST /api/v1/classification/bulk
Accept a list of `system_id` values, run classification on each, and return results per system. Handle partial failures gracefully — don't fail the whole batch if one system errors.

**File:** `backend/app/api/v1/classification.py`

Nileshkrishesh:
Closes #51

## Changes Made
- Added bulk classification endpoint
- Added support for multiple system_ids
- Implemented partial failure handling
- Added per-system response schema
- Updated API documentation

### DIFF
--- a/backend/app/api/v1/classification.py
+++ b/backend/app/api/v1/classification.py
@@ -16,7 +16,7 @@ router = APIRouter()
 class BulkClassificationItem(BaseModel):
     system_id: int
     classification: Optional[RiskClassificationResponse] = None
-    error: Optional[str] = None
+    error: Optional[str] = None 
 
 
 class BulkClassificationRequest(BaseModel):
@@ -272,3 +272,4 @@ def bulk_classify_systems(
         )
 
     return BulkClassificationResponse(results=results)
+    #done
--- a/backend/app/api/v1/classification.py
+++ b/backend/app/api/v1/classification.py
@@ -1,6 +1,8 @@
 from fastapi import APIRouter, Depends, HTTPException, status
 from sqlalchemy.orm import Session
-from typing import List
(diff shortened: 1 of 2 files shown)


SdSarthak (changes requested):
Solid feature with good partial-failure handling per system — missing questionnaire, not found, and invalid data all return per-system error entries rather than failing the whole batch. The integration test covers the mixed success/failure case correctly.

One issue: **`db.commit()` is called once after all systems are processed**, but the `assessment` inserts accumulate across the loop. If any `db.add(assessment)` fails mid-loop (e.g. a DB constraint), all previously processed systems in that batch roll back silently. Move `db.add(assessment)` and `db.commit()` inside a per-system `try/except` block so failures are isolated, consistent with the partial-failure design intent.

Also: `system.questionnaire_responses = system.questionnaire_responses` on line 221 is a no-op — remove it.

Nileshkrishesh (commented):
Thanks for the review! 

Updated the implementation by:
- Moving `db.add()` and `db.commit()` inside the per-system `try/except`
- Adding rollback handling for isolated failures
- Removing the redundant questionnaire response assignment

This ensures failures are isolated per system and do not affect successful records in the batch.

SdSarthak (approved):
All requested changes addressed:  and  are now inside their own  with  on failure, the no-op  line is removed, and each error path appends a  with an error message. LGTM.

SdSarthak:
Closing due to merge conflicts with recently merged changes to `ai_systems.py`. Please rebase onto `main` and reopen.

Issue #239, opened 2026-05-12
Create Standardized ML Training Pipeline for Safety Prediction System


divyankaaaaaa:
## Description
The current project lacks a structured and reproducible ML workflow for training and evaluating safety prediction models.

This issue proposes building a standardized machine learning pipeline to improve:
- reproducibility
- maintainability
- scalability
- model evaluation consistency

### 1. Dataset Organization
- Structured dataset folders
- Version-friendly format
- CSV validation checks

### 2. Data Preprocessing
- Missing value handling
- Feature encoding
- Feature scaling
- Train/test split pipeline

### 3. Training Pipeline
- Modular training scripts
- Configurable model selection
- Hyperparameter support

### 4. Evaluation System
- Accuracy
- Precision
- Recall
- F1-score
- Confusion matrix visualization

### 5. Model Persistence
- Save trained models using joblib/pickle
- Standardized inference loading

### 6. Inference API
- REST API endpoint for predictions
- JSON-based input/output

### 7. Documentation
- Setup instructions
- Training guide
- API usage examples

## Suggested Tech Stack
- Python
- Scikit-learn
- Pandas
- FastAPI / Flask
- Joblib

## Expected Outcome
A clean and reusable ML pipeline that allows contributors to:
- train models easily,
- evaluate performance consistently,
- deploy inference APIs reliably.

divyankaaaaaa:
Hey @SdSarthak, I would like to work on this issue under GSSoC 2026.

I plan to implement:
- dataset preprocessing pipeline
- modular training scripts
- evaluation framework
- inference API
- documentation for reproducibility

Please assign this issue to me.

bhaveshksahu:
Hi @SdSarthak, I'm a GSSoC '26 contributor specializing in Python and AI/ML. I have experience with FastAPI, Docker, and building multi-agent system architectures, and I would love to work on this issue. I’m particularly interested in streamlining ML workflows. Could you please assign it to me?

MaheshReddy-ML:
Hi @SdSarthak 👋
I have experience with Python and ML workflows, and I’d like to contribute to building the preprocessing, training, and evaluation pipeline for this issue. Kindly assign this issue to me under GSSoC'26.

Ankush-S803:
Hi @SdSarthak I have knowledge about Python and ML and FastApi so I'd like to contribute and work on this issue in GSSoC'26.Kindly assign this issue to me

vaishnavijha12:
[...]

PR #338, opened 2026-05-15
Create standardized ML training pipeline for Guard Safety Prediction …

MaheshReddy-ML:
Issue 239 Summary: Standardized ML Training Pipeline for Safety Prediction System(Under GSSoC'26)

What I have done

I created a proper training pipeline structure inside backend/app/modules/guard/training. Earlier, most of the training work was handled from one main train.py file. I separated that work into smaller folders so each part has a clear responsibility.

I added a config file at training/configs/training.yaml. This file keeps the dataset path, Hugging Face dataset name, split settings, model output path, training settings, artifact paths, and MLflow settings in one place.

I added data handling files under training/data. These files load the dataset, clean and standardize the prompt and label columns, map labels like 0 and 1 into benign and malicious, remove empty rows, remove duplicate prompts, and split the dataset into training and validation sets.

I added a trainer wrapper under training/trainer. It still uses the existing IntentClassifier.train method, so the actual model training behavior remains connected to the current guard classifier.

I added evaluation files under training/evaluation. These files calculate accuracy, weighted F1, macro F1, and a classification report. They also support saving predictions for later review.

I added pipeline entry points under training/pipelines. The train pipeline handles dataset loading, preprocessing, splitting, training, optional MLflow logging, and metrics artifact creation. The evaluation pipeline handles loading a dataset and model, running predictions, calculating metrics, and writing report artifacts.

I added utility files under training/utils for logging, saving JSON artifacts, creating run IDs, setting seeds, and optional MLflow logging.

I updated the new __init__.py files so they are not empty. They now expose the useful functions and classes from each folder. Where a module depends on heavier ML packages like torch, I used lazy loading so simple imports do not unnecessarily load the model stack.

I kept backend/app/modules/guard/train.py working as a compatibility wrapper. Existing commands and functions can still be used, but now they call the new standardized training modules internally.

I updated guard_config so the standard model folder is backend/app/modules/guard/models/classifier. I also kept the older models/intent_classifier path as a fallback so older trained models can still be detected.

I added .gitkeep files in the model and artifact folders so the expected folder structure remains in the repository before real checkpoints, metrics, or reports are generated.

I added PyYAML to backend/requirements.txt because the new training pipeline reads settings from a YAML config file.

Where they effect

These changes mainly affect the guard training area inside backend/app/modules/guard.

The existing runtime guard files like decision_engine.py, regex_rules.py, sanitizer.py, and llm_guard.py are not changed in behavior by this pipeline work.

The main affected training entry point is backend/app/modules/guard/train.py. It now delegates to the new pipeline structure, but it still supports the old download-only, train-only, and all-style workflow.

The model output location now points to backend/app/modules/guard/models/classifier as the standard location. The old backend/app/modules/guard/models/intent_classifier path is still supported as a fallback.

The training artifacts now have a standard place under backend/app/modules/guard/training/artifacts. Metrics, reports, and checkpoints can be stored there in a predictable way.

The new __init__.py exports make it easier for other files to import training helpers directly from the package folders instead of reaching into individual implementation files every time.

Why I have done

I made these changes because issue 239 asks for a standardized ML training pipeline for the safety prediction system.

The previous training setup worked, but too many responsibilities were grouped together in one file. Dataset downloading, preprocessing, splitting, training, and CLI behavior were all mixed together. That makes future updates harder and increases the chance of breaking something while changing a small part.

The new structure makes the system easier to maintain. If someone wants to update preprocessing, they can work in training/data. If they want to change evaluation metrics, they can work in training/evaluation. If they want to adjust the full workflow, they can work in training/pipelines.

I also kept backward compatibility because the existing project already depends on the old training script. Instead of removing that flow, I connected it to the new modules so existing usage keeps working.

This setup also prepares the project for future improvements like better metrics, experiment tracking, checkpoint management, different datasets, or additional model training strategies.

Thanks for giving me this all.


bhaveshksahu (approved):
Great work on this refactoring, Mahesh! The modular structure with the guard/train and guard/models directories makes the entire workflow much more professional and easier to follow.
I especially like the addition of guard_config.py—it’s going to make my job way easier when I start plugging in the different hyperparameters for the models.
I’ve looked through the files and everything looks solid and well-organized. Approved! ✅ >
(P.S. Still fighting with my laptop charger, but I’m ready to drop the logic into those new folders the second I’m back online!

MaheshReddy-ML:
> Great work on this refactoring, Mahesh! The modular structure with the guard/train and guard/models directories makes the entire workflow much more professional and easier to follow. I especially like the addition of guard_config.py—it’s going to make my job way easier when I start plugging in the different hyperparameters for the models. I’ve looked through the files and everything looks solid and well-organized. Approved! ✅ > (P.S. Still fighting with my laptop charger, but I’m ready to drop the logic into those new folders the second I’m back online!

Thanks for reviewing it @bhaveshksahu

MaheshReddy-ML:
@SdSarthak can you please label it as GSSoC approved  or GSSoC

MaheshReddy-ML:
@SdSarthak Hey , I guess you missed to close the issue #239

Issue #325, opened 2026-05-15
bug(frontend): DocumentEditor silent save failure — no UI feedback when save or auto-save fails


hemish22:
## Bug Description

When a document save fails in `DocumentEditor.tsx`, the error is only logged to the console. The user sees the saving spinner disappear with no indication that their changes were lost.

```ts
} catch (error) {
  console.error('Save failed:', error)
}
```

This applies to both the manual Save button and the auto-save triggered after 2 seconds of inactivity.

## Steps to Reproduce

1. Open a document in the editor
2. Disconnect from the backend (stop the backend container)
3. Make an edit and wait 2 seconds (auto-save triggers)
4. No error shown — user believes changes are saved

## Expected Behaviour

An error message is displayed in the UI when save fails so the user knows their changes were not persisted.

## Fix

Add an error state to the component and display it in the UI when the save request fails.

## Files Affected

- `frontend/src/components/DocumentEditor.tsx` lines 41–43

hemish22:
i am a gssoc 2026 contributor and will be working on this issue

higgsboson1710:
@SdSarthak   Could u plz assign this to me i got the  necessary knowledge  to work on this

srinivasareddykovvuri116-dot:
Hi @SdSarthak!, I'd like to work on this issue under GSSoC 2026 if it is still available.

My plan is to add an error state in `DocumentEditor.tsx` so that failed manual saves and auto-save operations display clear UI feedback instead of only logging errors to the console. I’ll ensure the error message appears when the backend request fails and clears appropriately after a successful save.

This should improve user awareness and prevent silent data-loss confusion.

Please assign this issue to me if available.

divya-123-TC:
Hi @SdSarthak,

I would like to work on this issue under GSSoC'26.

I reviewed the issue and understood the problem related to silent save failures in `DocumentEditor.tsx`. Currently, when manual save or auto-save fails, the error is only logged in the console, which can mislead users into thinking their changes were saved successfully.

Proposed Solution:

* Add a dedicated error state inside the `DocumentEditor` component
* Display a clear and user-friendly UI error message whenever save or auto-save fails
* Ensure the error message appears consistently for both manual save and auto-save operations
* Clear the error state automatically after a successful save attempt
* Maintain proper responsiveness and visual consistency with the existing editor UI

I will make sure the implementation is clean, properly tested, and aligned with the existing frontend structure before submitting the PR.

[...]

Issue #324, opened 2026-05-15
bug(frontend): XSS vulnerability in DocumentEditor — marked() output rendered without sanitization


hemish22:
## Bug Description

`DocumentEditor.tsx` renders markdown preview using `dangerouslySetInnerHTML` with unsanitized output from `marked()`. This creates a Cross-Site Scripting (XSS) vulnerability — malicious HTML in document content executes in the user's browser.

```tsx
<div dangerouslySetInnerHTML={{ __html: marked(content) }} />
```

`marked` converts markdown to HTML but does not sanitize it. Tags like `<script>`, `<img onerror=...>`, and `<a href="javascript:...">` will execute.

## Steps to Reproduce

1. Open any document in the editor
2. Insert the following into the content:
   ```
   <img src=x onerror="alert('XSS')">
   ```
3. Switch to Preview mode
4. Alert executes in the browser

## Expected Behaviour

HTML in document content is sanitized before rendering. Malicious tags are stripped.

## Fix

Install `dompurify` and wrap the output:
```tsx
import DOMPurify from 'dompurify'
<div dangerouslySetInnerHTML={{ __html: DOMPurify.sanitize(marked(content) as string) }} />
```

## Files Affected

- `frontend/src/components/DocumentEditor.tsx` line 104

hemish22:
i am a gssoc 2026 contributor and will be working on this issue

saurabhhhcodes:
Hi @SdSarthak, I would like to work on this security bug under GSSoC.

Plan:
- replace raw `marked()` preview output with a sanitized rendering path using a well-maintained sanitizer such as DOMPurify
- block dangerous tags/attributes and unsafe URL schemes like `javascript:` while preserving normal markdown formatting
- add regression coverage for `<img onerror>`, `<script>`, and unsafe links so this cannot regress silently
- keep the fix scoped to `DocumentEditor` and the existing markdown preview flow
- verify the frontend locally after the patch

This is a user-facing XSS issue, so I’ll keep the PR focused and security-oriented. Please assign this to me if available.

saurabhhhcodes:
Linked fix here: PR #593

saurabhhhcodes:
Correction: linked fix is PR #668

saurabhhhcodes:
Current main now already includes DocumentEditor preview sanitization via DOMPurify, so PR #668 is no longer a distinct fix branch.

saurabhhhcodes:
Linked fix here: PR #677

PR #359, opened 2026-05-16
fix(frontend): fix DocumentEditor auth header, XSS, and silent save failure

### LINKED ISSUE #323: bug(frontend): DocumentEditor uses raw fetch without Authorization header — all document saves return 401
## Bug Description

`DocumentEditor.tsx` uses a raw `fetch` call instead of the existing axios instance from `services/api.ts`. This means the Authorization Bearer token is never attached to the request. Every document save silently fails with a 401 Unauthorized response.

The code even has a comment admitting the problem:
```ts
// Add auth header if needed - check how other API calls are done in your app
```

## Steps to Reproduce

1. Login and open any generated document
2. Edit the content
3. Click Save or wait for auto-save
4. Check browser network tab — PUT `/api/v1/documents/{id}` returns 401

## Expected Behaviour

Document saves successfully using the authenticated user's JWT token.

## Root Cause

`frontend/src/components/DocumentEditor.tsx` line 29 uses `fetch()` directly.
All other API calls use the `api` axios instance from `frontend/src/services/api.ts` which attaches the token via interceptor.

## Fix

Replace the `fetch` call with the `api` axios instance from `services/api.ts`.

## Files Affected

- `frontend/src/components/DocumentEditor.tsx`

hemish22:
## Summary

- **Auth (#323):** Replaced raw `fetch()` with the `api` axios instance from `services/api.ts`. The interceptor now attaches the `Authorization: Bearer <token>` header automatically, so document saves no longer return 401.
- **XSS (#324):** Wrapped `marked()` output with `DOMPurify.sanitize()` before passing to `dangerouslySetInnerHTML`. Malicious HTML in document content (e.g. ``, `<script>`) is now stripped before rendering.
- **Silent save failure (#325):** Added `saveError` state. When a save request fails, an error message is shown in the editor toolbar so users know their changes were not persisted.

## Files Changed

- `frontend/src/components/DocumentEditor.tsx` — all three fixes
- `frontend/package.json` / `package-lock.json` — added `dompurify` and `@types/dompurify`

## Test Plan

- [ ] Login, open a generated document, edit content — save should succeed (check network tab for 200, not 401)
- [ ] In preview mode, paste `` — alert should NOT fire
- [ ] Stop backend, edit document — red error message should appear in toolbar

Closes #323, #324, #325


hemish22:
Closing in favour of 3 separate PRs — one per issue for clearer review.

PR #360, opened 2026-05-16
fix(frontend): use axios instance in DocumentEditor to attach auth token

### LINKED ISSUE #323: bug(frontend): DocumentEditor uses raw fetch without Authorization header — all document saves return 401

hemish22:
## Summary

`DocumentEditor.tsx` used a raw `fetch()` call with no `Authorization` header, so every document save returned `401 Unauthorized`. This replaces the `fetch` with the existing `api` axios instance from `services/api.ts`, which attaches the `Bearer` token automatically via its request interceptor.

## Root Cause

`fetch()` has no knowledge of the app's auth state. The `api` axios instance already has a request interceptor that reads the token from Zustand store and sets `Authorization: Bearer <token>` on every request.

## Changes

- `frontend/src/components/DocumentEditor.tsx` — replace `fetch` with `api.put`; import `api` from `../services/api`

## Test Plan

- [ ] Login, open a generated document, edit content, click Save
- [ ] Check network tab — `PUT /api/v1/documents/{id}` should return `200`, not `401`

Closes #323

### DIFF
--- a/frontend/src/components/DocumentEditor.tsx
+++ b/frontend/src/components/DocumentEditor.tsx
@@ -3,6 +3,7 @@ import { Save, Eye, EyeOff } from 'lucide-react'
 import CodeMirror from '@uiw/react-codemirror'
 import { markdown } from '@codemirror/lang-markdown'
 import { marked } from 'marked'
+import api from '../services/api'
 
 interface DocumentEditorProps {
   documentId: number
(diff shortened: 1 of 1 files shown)


SdSarthak:
Closing as duplicate — the same auth fix (replacing raw `fetch` with the project's axios `api` instance in `DocumentEditor.tsx`) was applied via #374.

PR #362, opened 2026-05-16
fix(frontend): show error message in toolbar when document save fails

### LINKED ISSUE #325: bug(frontend): DocumentEditor silent save failure — no UI feedback when save or auto-save fails

hemish22:
## Summary

When a document save failed in `DocumentEditor.tsx`, the error was only logged to `console.error`. The saving spinner disappeared with no UI feedback — users had no way to know their changes were lost.

This adds a `saveError` state and displays a red error message in the editor toolbar whenever a save request fails (both manual save and auto-save).

## Reproduction

1. Open a document, make an edit
2. Stop the backend (`docker stop aegisai_backend`)
3. Wait 2 seconds for auto-save, or click Save
4. No error shown — user believes save succeeded (**before** this fix)

## Changes

- `frontend/src/components/DocumentEditor.tsx` — add `saveError` state; set on network failure or non-ok response; render in toolbar

## Test Plan

- [ ] Stop backend, edit document, wait for auto-save — red "Save failed" message appears in toolbar
- [ ] Successful save — no error message shown

Closes #325

### DIFF
--- a/frontend/src/components/DocumentEditor.tsx
+++ b/frontend/src/components/DocumentEditor.tsx
@@ -20,10 +20,12 @@ export default function DocumentEditor({
   const [content, setContent] = useState(initialContent)
   const [showPreview, setShowPreview] = useState(false)
   const [isSaving, setIsSaving] = useState(false)
+  const [saveError, setSaveError] = useState<string | null>(null)
   const [saveTimeout, setSaveTimeout] = useState<ReturnType<typeof setTimeout> | null>(null)
 
   const handleSave = useCallback(async () => {
     setIsSaving(true)
+    setSaveError(null)
     // Call PUT endpoint
     try {
       const response = await fetch(`/api/v1/documents/${documentId}`, {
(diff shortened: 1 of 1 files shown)


SdSarthak:
The `saveError` state and toolbar display idea is good, but this PR now conflicts with #374 (just merged) which replaced the `fetch` call with the axios `api` instance and restructured `handleSave`. The error state should be added on top of the merged auth fix — the approach is sound, but needs a rebase onto `main` before it can be applied.

PR #383, opened 2026-05-16
Add skip and limit pagination to list endpoints

### LINKED ISSUE #4: Add skip/limit pagination query params to list endpoints
Add `skip: int = 0` and `limit: int = 20` query parameters to `GET /api/v1/ai-systems` and `GET /api/v1/documents`. Apply them in the SQLAlchemy query with `.offset(skip).limit(limit)`.

**Tip:** Use the `PaginatedResponse` generic schema once that issue is merged.

sriharshitha128:
## Summary

Closes #4

Added pagination support to the list endpoints by implementing `skip` and `limit` query parameters.

Updated endpoints:

* GET /api/v1/ai-systems
* GET /api/v1/documents

Implemented pagination using SQLAlchemy `offset()` and `limit()` methods and verified the changes successfully in Swagger UI.

### DIFF
--- a/backend/app/api/v1/ai_systems.py
+++ b/backend/app/api/v1/ai_systems.py
@@ -50,22 +50,38 @@ _SORTABLE_FIELDS = {
     "created_at": AISystem.created_at,
 }
 
-
 @router.get("/", response_model=PaginatedResponse[AISystemResponse])
 def list_ai_systems(
-    sort_by: Optional[str] = Query("created_at", description="Sort field: name, risk_level, compliance_score, created_at"),
-    order: Optional[str] = Query("desc", description="Sort direction: asc, desc"),
-    page: int = Query(1, ge=1, description="Page number (1-indexed)"),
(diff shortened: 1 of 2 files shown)


SdSarthak:
Two issues prevent merging this:

1. **Breaking import removal**: `documents.py` uses `StreamingResponse` in the document export endpoint, but this PR removes the `from fastapi.responses import FileResponse, StreamingResponse` import line. Merging would cause an immediate `NameError` at runtime.

2. **Breaking API change**: Switching from page-based (`page=1&limit=50`) to skip-based (`skip=0&limit=20`) pagination changes the query parameter contract. The frontend currently sends `page` and the existing `PaginatedResponse` schema returns a `page` field that callers depend on. This would need a coordinated frontend update and a migration plan.

The formatting/whitespace cleanup around the Query parameters is fine on its own. If skip-based pagination is the right direction long-term, it would need its own issue/discussion and coordinated frontend changes.

PR #389, opened 2026-05-16
refactor: align report generator with existing RAG pipeline

### LINKED ISSUE #81: Build LLM-powered narrative compliance report generation
Replace the current `.format()` document templates with LLM-generated prose. Given a system's metadata, risk classification, and questionnaire answers, use the LLM + RAG context to write a full narrative compliance report grounded in the actual regulation text.

The output should read like a professional compliance document written by a legal expert, not a filled-in template.

avniuraon:
## Summary

Closes #81

Refactored the report generator to align with the existing FAISS/LangChain RAG pipeline. Removed the OpenAI dependency and replaced custom embedding/vector store logic with the project’s retriever and LLM client.

## Type of Change

- [ ] Bug fix
- [x] Refactor
- [ ] New feature

## Notes

- Removed OpenAI API usage
- Integrated existing retriever for context retrieval
- Uses project LLMClient for response generation
- Avoids startup issues due to external dependencies

### DIFF
--- a/backend/app/api/v1/ai_systems.py
+++ b/backend/app/api/v1/ai_systems.py
@@ -353,3 +353,28 @@ def get_ai_system_history(
         }
         for log in history
     ]
+
+@router.post("/{system_id}/generate-report")
+def generate_report_endpoint(
+    system_id: int,
+    db: Session = Depends(get_db),
+    current_user: User = Depends(get_current_user),
+):
+    system = db.query(AISystem).filter(
+        AISystem.id == system_id,
+        AISystem.owner_id == current_user.id
+    ).first()
+
+    if not system:
+        raise HTTPException(status_code=404, detail="System not found")
+
+    from app.modules.rag.report_generator import generate_report
+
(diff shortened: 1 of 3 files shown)


avniuraon:
Hi! I’ve updated this based on the previous feedback — removed the OpenAI dependency and aligned the implementation with the existing FAISS/LangChain RAG pipeline. Would appreciate another review!

SdSarthak:
The `generate_report` function signature requires a `retriever` parameter (`def generate_report(metadata, risk, answers, retriever)`) but the endpoint calls it without one: `generate_report(metadata=..., risk=..., answers=...)`. This would raise a `TypeError` at runtime on every POST to `/{system_id}/generate-report`. Additionally, a compliance report generator for a specific AI system already exists via the documents endpoint — this endpoint would need to be reconciled with that flow rather than added as a parallel path. Please fix the missing retriever argument and clarify how this relates to the existing document generation endpoint before resubmitting.

PR #427, opened 2026-05-17
Deploy application and add live demo link

adityakumar9876:
Successfully deployed the application and verified the live site.

Live Demo: https://aegisai-gg0v.onrender.com/login


SdSarthak:
Two issues:

1. **Removes TypeScript type checking from the build**: Changing `"build": "tsc && vite build"` to `"build": "vite build"` means TypeScript errors will no longer fail the build. Type errors would silently ship to production. The correct fix if there are type errors blocking deploy is to fix them, not skip the check.

2. **Removes ThemeToggle import**: Deletes the import without explanation — if ThemeToggle doesn't exist yet, this might be intentional, but it should come with context.

The `vite.config.ts` additions (host: true, allowedHosts: ['.onrender.com']) are fine for deployment configuration. Please resubmit with just those vite config changes and restore the TypeScript build step.

adityakumar9876:
Thank you for the review and feedback.

The additional changes to package.json and ThemeToggle were made while troubleshooting deployment issues on Render. I understand that these changes were unrelated to the deployment fix and should not have been included in the PR.

I have now corrected the deployment issues locally, restored the original TypeScript build step, and removed the duplicate ThemeToggle import(duplicate ThemeToggle import was causing issue in rendering). I will submit a clean PR containing only the required vite.config.ts update,Single ThemeToggle import and the live demo link.

Thanks again for the guidance.

PR #435, opened 2026-05-18
fix(tests): add get_current_user override in conftest to fix RAG test 401s

### LINKED ISSUE #393: bug(ci): Frontend build fails on Node 18 and backend tests return 401 due to missing auth mock

Thongramdhanapyari:
Closes #393

## Changes
- Added `get_current_user` dependency override in the test client fixture in conftest.py
- This fixes 7 RAG tests in test_rag_ingest.py that were returning 401 Unauthorized

## Proof
On main: 7 RAG tests fail with 401, 126 passing
After fix: 7 RAG tests pass, 128 passing

## Type of Change
- [x] Bug fix

## GSSoC '26
I am a GSSoC '26 contributor (@Thongramdhanapyari)

### DIFF
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -49,7 +49,14 @@ def client(db_engine):
     def override_get_db():
         yield session
     
+    from app.api.v1.auth import get_current_user
+    from app.models.user import User
+
+    def override_get_current_user():
+        return User(id=1, email="test@example.com", is_active=True, hashed_password="fake")
+
     app.dependency_overrides[get_db] = override_get_db
+    app.dependency_overrides[get_current_user] = override_get_current_user
     
     client = TestClient(app)
     yield client


Thongramdhanapyari:
Note: Both CI failures are pre-existing on main and unrelated to this PR which only touches backend/tests/conftest.py. The frontend failure is the d3 package-lock issue and the backend failure includes the same 5 pre-existing test failures. The 7 RAG tests that this PR fixes are now passing.

SdSarthak:
The approach is correct — adding a `get_current_user` override to conftest is exactly the right fix. However the import should be `from app.core.security import get_current_user` (not `from app.api.v1.auth import get_current_user`). FastAPI's dependency override system works by object identity: since all endpoints import `get_current_user` from `app.core.security`, overriding the version imported from `app.api.v1.auth` (which is a different name reference) won't affect those endpoints. Please fix the import and resubmit.

PR #495, opened 2026-05-21
fix: prevent information exposure through verbose exception messages

### LINKED ISSUE #226: Fix Information Exposure via Verbose Exception Messages
### Describe the bug

Several API endpoints in the backend currently return raw exception strings to the client in the `detail` field of an `HTTPException`. 

For example, in `backend/app/api/v1/guard.py`:
```python
except Exception as e:
    raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=str(e))
```

### Impact

This is a security vulnerability (Information Exposure). Raw exception strings can leak sensitive internal information, including:
- Local file system paths
- Database schema details or connection issues (potentially revealing credentials)
- Internal variable names and logic structure

### Expected behaviour

The API should log the full stack trace internally for debugging but return a generic, safe error message to the client (e.g., "An internal error occurred while processing the Guard scan").

### Steps to reproduce

1. Cause a controlled failure in a module (e.g., rename a model file so the Guard module fails to load).
2. Call the affected endpoint.
3. Observe the full, descriptive error message in the API response.

### AegisAI version / commit
0.1.x (current)

Aryashjain12:
## Summary

Closes #226

This PR fixes an information exposure vulnerability in the Guard API endpoints caused by returning raw exception strings directly to API clients using `detail=str(e)`.

The update adds internal exception logging using Python's logging module while returning sanitized generic HTTP 500 responses to users. The fix has been applied consistently to both `/guard/scan` and `/guard/scan/batch` endpoints.

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Documentation update
* [ ] Refactor
* [ ] Tests
* [ ] Infra / CI

## Checklist

* [x] I have read [[CONTRIBUTING.md](https://chatgpt.com/CONTRIBUTING.md)](../CONTRIBUTING.md)
* [x] My code follows the project style (PEP 8 for Python, ESLint for TS)
* [ ] I have added/updated tests where relevant
* [ ] `pytest backend/tests/` passes locally

## Screenshots (if UI change)

N/A

### DIFF
--- a/backend/app/api/v1/guard.py
+++ b/backend/app/api/v1/guard.py
@@ -13,6 +13,7 @@ import hashlib
 from collections import defaultdict, deque
 from datetime import datetime, timedelta, timezone
 from threading import Lock
+import logging
 
 from fastapi import APIRouter, Depends, HTTPException, Query, status
 from fastapi.responses import JSONResponse
@@ -29,7 +30,7 @@ from app.schemas.guard_scan_log import GuardScanLogResponse
 from app.schemas.pagination import PaginatedResponse
 
 router = APIRouter()
-
+logger = logging.getLogger(__name__)
 
 _RATE_LIMIT_REQUESTS = 60
 _RATE_LIMIT_WINDOW_SECONDS = 60
@@ -145,8 +146,11 @@ def scan_prompt(
             ),
         )
     except Exception as e:
+        logger.exception("Guard scan failed")
+
         raise HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=str(e)
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="An internal error occurred while processing the Guard scan."
         )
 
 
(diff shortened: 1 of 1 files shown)


SdSarthak (commented):
The change is exactly right — swapping `detail=str(e)` for a generic user-facing message while logging the real exception internally is good security hygiene. The same pattern should also be applied to the third occurrence at line ~366 in the bulk scan endpoint.

The PR currently conflicts with recent merges (#480, #490) that rewrote the bulk scan error path. Please rebase onto main and re-check — the logic itself is correct and will merge cleanly once rebased.

Aryashjain12:
Hi @SdSarthak — rebased onto main after #480 and #490, conflicts resolved.
To address your review note on the third occurrence at ~line 366: that location in the post-merge file contains the db.rollback() added by #480. I've preserved it and added logger.exception("Bulk guard scan failed") alongside it, so the bulk scan endpoint now rolls back the transaction, logs the full stack trace internally, and returns the sanitized generic message to the client — consistent with the single scan endpoint.
Both except Exception blocks are now fixed. The except ValueError block at line ~283 uses detail=str(e) intentionally — it catches a hardcoded ValueError from validate_prompts(), so the message is always the controlled string "Maximum 50 prompts allowed per batch request.", not any system internals.
Ready for re-review.

PR #540, opened 2026-05-23
Feature/issue 88 regulation ruleset plugins

### LINKED ISSUE #88: Add plugin system for custom regulation schemas — YAML-based rulesets
Design a YAML-based schema for defining regulation rulesets loadable at runtime without modifying core Python. Each ruleset defines: regulation name, risk factors, prohibited uses, required documents, compliance questions.

Ship EU AI Act as the built-in ruleset. Contributors can add UK AI Bill, India DPDP, etc. as YAML files.

This is the architecture prerequisite for multi-regulation support.

EHEGUY:
## Description
closes #88. 

[Provide a brief summary of what this feature does, why it was added, and any important implementation details the reviewer should know about.]

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactor
- [ ] Tests
- [ ] Infra / CI

## Checklist

## Screenshots (if UI change)

*N/A - Backend only change.

singleton answer: Yes, module-level singleton initialized in the FastAPI lifespan event is the right call. Per-request loading would hit disk on every compliance check wasteful and fragile. The singleton loads once at startup, fails fast on bad YAMLs before the server ever accepts traffic, and is trivially replaceable later with a hot-reload watcher.

The maps_to integrity check in test (a) is worth keeping  it's a cheap guard that prevents a YAML contributor from referencing a non-existent risk factor ID in their questions.

Custom override behavior (test h) is the correct UX for contributors who want to patch a built-in ruleset regionally without forking the repo.*


SdSarthak:
This PR bundles three unrelated features: regulation ruleset plugins, webhook API delivery, and classification UI tabs. Each should be a separate PR so they can be reviewed and merged independently.

Additionally, the webhooks.py routes use `async def` with a synchronous SQLAlchemy `Session`, which will block the event loop — the same issue noted in the related PR. Please split the regulation loader into its own PR and keep the async/sync issue in mind for the webhooks rework.

PR #543, opened 2026-05-23
 issue441: audit logging

### LINKED ISSUE #441: Add audit logging for Guard scan decisions
### Motivation

Currently, Guard scan decisions (allow/block/sanitize) are not persistently logged. This makes it difficult to track security-related events, review moderation history, and support compliance/audit requirements under AI governance frameworks such as the EU AI Act.

Adding audit logs would improve observability, traceability, and debugging for the Guard module while enabling future analytics and compliance reporting features.

### Proposed Solution

Implement persistent audit logging for every Guard scan request.

The feature would include:

Creating an audit_log database model/table
Logging:
user ID
scan decision (allow/block/sanitize)
timestamp
confidence/risk score if available
optional prompt metadata or hashed prompt
Automatically storing logs whenever /guard/scan is called
Adding unit/integration tests for audit log creation and retrieval
Adding migration support using Alembic

LLM Guard

### Alternatives Considered

Logging only to console output
Using external monitoring tools without database persistence
Storing only aggregated metrics instead of per-request audit records

These alternatives would not provide the same level of traceability or compliance support as persistent structured audit logs.

Anshuman-cs50:
## Description
This PR implements persistent audit logging for the LLM Guard module to improve observability, traceability, and aid in supporting compliance/audit requirements under AI governance frameworks (e.g., the EU AI Act).

Previously, Guard scan decisions were ephemeral. With this change, every scan request is persistently logged in the database, capturing critical security and moderation metrics. 

## Changes Made
* **Audit Log Database Model:** Utilized the `GuardScanLog` database model to log user IDs, hashed prompts (SHA-256 to protect user privacy), scan decisions (allow/block/sanitize), confidence scores, and matched patterns.
* **Scan Event Logging:** Updated `app/api/v1/guard.py` to automatically store a `GuardScanLog` whenever `/guard/scan` or `/guard/scan/batch` is invoked.
* **Alembic Migrations:** Autogenerated an Alembic migration for the `guard_scan_logs` table to ensure seamless deployment and schema management.
* **Testing:** Added robust integration tests in `tests/test_guard_api.py` to verify audit log creation during a prompt scan and to test the retrieval and pagination via the history endpoint.

## Related Issues
* Closes #441 

## Testing Performed
* `pytest tests/test_guard_api.py` - Validated log generation and retrieval functionality.
* Created a test local SQLite database and confirmed that the generated Alembic migration successfully applies.


SdSarthak:
The migration in this PR recreates all existing tables (`users`, `ai_systems`, `guard_scan_logs`, `notifications`, `documents`, etc.) instead of only adding the new table. This would fail with "table already exists" on any database that has already been migrated. The migration was auto-generated without first applying the existing migrations, so Alembic included all tables in scope.

To fix: apply all existing migrations first (`alembic upgrade head`), then regenerate with `alembic revision --autogenerate -m "add_guard_scan_logs"`. The generated file should only contain DDL for tables that don't already exist.

PR #550, opened 2026-05-23
Fix NotificationBell API integration and unread handling

### LINKED ISSUE #486: Wire NotificationBell to real notifications API and unread polling
### Motivation

The bell should reflect real unread data from the backend, but it is not fully wired to the API flow. That means users cannot reliably see unread counts or preview recent notifications from the header.

### Proposed Solution

NotificationBell in NotificationBell.tsx is still using stubbed assumptions and needs to be connected to the real notifications endpoints.

Scope:
- Fetch unread notifications from GET /api/v1/notifications?unread_only=true
- Poll every 60 seconds so the count stays current
- Show the unread badge in the header
- Display the 5 most recent notifications in the dropdown
- Mark a notification as read when clicked and navigate to the notifications page
- Handle loading and empty states cleanly

I want to work on this feature under GSSoC 2026.

Frontend

### Alternatives Considered

_No response_

jainiksha:
Closes #486

## What does this PR do?

This PR fixes the NotificationBell API integration and unread notification handling.

### Changes Made
- Connected NotificationBell with backend notifications API
- Fixed white screen crash caused by invalid notifications handling
- Added unread notifications count support
- Limited dropdown to latest 5 notifications
- Added safe array handling for notifications
- Improved empty state rendering

## Type of Change
- [x] Bug fix
- [x] Refactor

## Checklist
- [x] My code follows the project style
- [x] I have tested the changes locally
- [x] I have not committed secrets or .env files


SdSarthak:
Three issues:

1. **backend/package-lock.json**: A Node.js package lock file committed into the Python backend directory. The backend uses pip/requirements.txt, not npm. Please remove this file.

2. **.env.example**: Changes `SECRET_KEY` from an intentional placeholder comment to `SECRET_KEY=mysecretkey`. An example file with a real-looking secret value encourages developers to use it as-is, which is a security risk. Please revert to the comment form.

3. **Indentation**: In the NotificationBell fix, `.slice(0, 5).map(...)` is dedented to column 2 instead of staying aligned with the surrounding code.

The actual fix — adding `Array.isArray()` guards to handle the paginated response shape — is the right approach and needed. Please re-submit with only that change, fixing the indentation.

PR #554, opened 2026-05-23
test: add unit tests for PaginatedResponse schema

### LINKED ISSUE #100: Write unit tests for PaginatedResponse generic schema
Add tests in `backend/tests/test_schemas.py` for the `PaginatedResponse[T]` schema (once that issue is merged):
- Assert `items`, `total`, `page`, `limit` fields are present
- Assert `items` is correctly typed as a list
- Assert schema serialises and deserialises correctly with different item types
- Assert `total` can be larger than `len(items)` (valid paginated subset)

**Depends on:** PaginatedResponse schema issue

0975-coder:
## Description

Added unit tests for the `PaginatedResponse[T]` generic schema.

### Tests Added

* Verified required fields exist
* Verified `items` is a list
* Verified serialization/deserialization
* Verified `total` can exceed current item count

Fixes #100


SdSarthak:
The import path `from backend.schemas.pagination import PaginatedResponse` is wrong — pytest runs from the `backend/` directory, so the correct path is `from app.schemas.pagination import PaginatedResponse`. With the current import this test will fail with a ModuleNotFoundError at collection time. Please fix the import path.

PR #564, opened 2026-05-23
fix: add bcrypt password length validation

### LINKED ISSUE #419: Bug Report: Handle bcrypt Password Length Limitation and Improve Authentication Validation
### Describe the bug

Authentication-related backend tests fail when passwords exceed bcrypt’s supported length limit.

The current implementation allows overly long passwords to reach the bcrypt hashing/verification layer, resulting in runtime exceptions such as:

```plaintext
ValueError: password cannot be longer than 72 bytes
```

Additionally, some authentication flows return unexpected `401 Unauthorized` responses instead of consistent validation errors, indicating inconsistent authentication validation and error handling behavior.

### Steps to reproduce

1. Go to the authentication or user registration endpoint.
2. Submit a password longer than 72 bytes.
3. Call the authentication/login endpoint using the same credentials.
4. Observe runtime exceptions or inconsistent `401 Unauthorized` responses.

### Expected behaviour

* Password length should be validated before bcrypt hashing or verification occurs.
* Requests with overly long passwords should return clean validation errors.
* Authentication flows should fail gracefully and consistently.
* Authentication-related tests should behave predictably without uncaught exceptions.

### Actual behaviour

* bcrypt raises runtime exceptions for passwords longer than 72 bytes.
* Some authentication flows return inconsistent `401 Unauthorized` responses.
* Authentication validation behavior is not fully defensive or standardized.
* Backend authentication tests fail unpredictably.

### AegisAI version / commit

Affected areas include:  * `backend/app/core/security.py` * authentication validation flow * backend authentication tests

24r01a6684:
## Description

Added password length validation before bcrypt hashing and verification to prevent runtime exceptions when passwords exceed bcrypt's 72-byte limit.

## Changes Made

* Added validation in `get_password_hash()`
* Added validation in `verify_password()`
* Prevents bcrypt runtime crashes for long passwords

## Type of Change

* [x] Bug fix

## Issue

Fixes #419

### DIFF
--- a/backend/app/core/security.py
+++ b/backend/app/core/security.py
@@ -28,29 +28,42 @@ def _get_credentials_exception() -> HTTPException:
 
 def verify_password(plain_password: str, hashed_password: str) -> bool:
     """Verify a plain password against its hashed version."""
+
+    if len(plain_password.encode("utf-8")) > 72:
+        raise ValueError("Password cannot be longer than 72 bytes for bcrypt.")
+
     return pwd_context.verify(plain_password, hashed_password)
 
 
 def get_password_hash(password: str) -> str:
     """Securely hash a password using bcrypt."""
+
+    if len(password.encode("utf-8")) > 72:
+        raise ValueError("Password cannot be longer than 72 bytes for bcrypt.")
+
     return pwd_context.hash(password)
 
 
 def create_access_token(data: dict, expires_delta: Optional[timedelta] = None) -> str:
     """Create a JWT access token with an expiration payload."""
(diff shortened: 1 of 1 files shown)


SdSarthak:
This fix is already in main — the 72-byte bcrypt validation was merged earlier via #451. Both `verify_password` and `get_password_hash` already raise on passwords longer than 72 bytes. No changes needed.

Also, please use a dedicated feature branch rather than submitting PRs from your fork's `main` — that makes it difficult to iterate on feedback.

PR #570, opened 2026-05-24
feat: add pagination controls to Documents list (#14)

Bhavy12-cell:
Closes #14

Added Prev/Next buttons and a page indicator to the Documents page.

Changes:
- `api.ts`: updated `documentsApi.list()` to accept `skip` and `limit` params
- `Documents.tsx`: added pagination state, wired query to skip/limit, added Prev/Next UI with page indicator, resets to page 1 on filter change

PR #571, opened 2026-05-24
feat: add pagination controls to Documents list (#14)

### LINKED ISSUE #14: Add pagination controls to Documents list (frontend)
Add Prev / Next buttons and a page indicator to the Documents page. Wire to `skip`/`limit` query params.

**Files:** `Documents.tsx`, `src/services/api.ts`

Bhavy12-cell:
Closes #14

Added pagination to the Documents list so users can navigate 
through documents page by page instead of loading everything at once.

- Shows 10 documents per page
- Prev/Next buttons to navigate
- Displays current page and total count
- Page resets to 1 when search or filters change

### DIFF
--- a/frontend/src/pages/Documents.tsx
+++ b/frontend/src/pages/Documents.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useState, useEffect } from 'react'
 import { useQuery, useMutation, useQueryClient } from '@tanstack/react-query'
 import { aiSystemsApi, documentsApi } from '../services/api'
 import { FileText, Download, Trash2, Plus, Edit, Copy, Check } from 'lucide-react'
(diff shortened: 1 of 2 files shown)


agsaru (commented):
This is duplicate pr of #570 . You pr description does not matches with changes.

SdSarthak:
Two issues with this implementation:

1. **Side effect during render**: `if (apiTotal !== total) setTotal(apiTotal)` is called directly in the render body (not inside a `useEffect`). Calling `setState` during render is a React anti-pattern — it triggers an immediate re-render on every cycle where `apiTotal !== total`, causing an infinite update loop or at minimum unexpected behaviour. Move this into a `useEffect([documentsData])` block.

2. **Inconsistent indentation**: The new state declarations (`const LIMIT`, `const [page]`, `const [total]`) and the `useQuery` block are dedented to column 1/2 instead of matching the surrounding 2-space indentation, and the `useEffect` above them calls `setPage` before `page` is declared (hoisting works for `let`/`var` but this reads as confusing).

The pagination concept (resetting page on filter change, showing total count) is right. Please fix the two points above.

Issue #335, opened 2026-05-15
BUG : Guard classifier using random weights


Hell1213:
**Describe the bug**

The Guard module loads microsoft/deberta-v3-small but the classification head has random weights since no fine-tuned model is included. This means prompt injection detection is essentially random.

**Steps to reproduce**

1. Fresh install of AegisAI
2. Call Guard API with test prompts
3. Try known injection patterns
4. Observe inconsistent/random blocking decisions

**Expected behaviour**

Guard should detect prompt injections reliably using a fine-tuned model

**Actual behaviour**

Model has untrained classification head, so decisions are unreliable. The training notebook exists but hasn't been executed.

**AegisAI version / commit**

main

Hell1213:
hey @SdSarthak I'd like to work on this. The issue is that the training notebook exists but hasn't been run, so we're loading a model with random weights. I can execute the notebook on the safe-guard dataset, save the trained weights, and add a download script or include them in the repo. Please assign this to me.

palakjaiswal16:
Hi @SdSarthak,

I’d like to work on this issue under GSSoC’26.

I understand that the Guard classifier is currently loading `microsoft/deberta-v3-small` with an untrained classification head, which makes the prompt injection detection unreliable. I can help by training/fine-tuning the classifier on the safe-guard dataset, saving the trained weights properly, and updating the loading/download flow so the Guard module uses the fine-tuned model instead of random weights.

Please assign this issue to me if it’s available. Thanks!

Rumeth951:
Hi @SdSarthak! I'd love to contribute to this issue as well.

I can help by:
- Running the existing training notebook on the safe-guard dataset
- Saving and hosting the fine-tuned weights (e.g., via HuggingFace Hub)
- Adding a download script so the Guard module pulls the trained model instead of relying on random weights
- Adding a basic sanity check / unit test to catch untrained heads early

Please assign this to me if it's still available. Happy to coordinate with others who've already expressed interest!

MaheshReddy-ML:
Hi @SdSarthak , I would like to work on this issue under GSSoC'26 . Assign this issue to me.

PR #573, opened 2026-05-24
Fix/335 guard classifier random weights

MaheshReddy-ML:
Guard Classifier Random Weights Bug Fix Summary #335 
==============================================

Issue
-----
The maintainer reported that the Guard module loaded microsoft/deberta-v3-small
directly when no fine-tuned classifier checkpoint was present. Because the code
used AutoModelForSequenceClassification with num_labels=3 on the base DeBERTa
model, HuggingFace created a new classification head with random weights.

That made prompt injection classification unreliable. The Guard API could return
inconsistent allow, sanitize, or block decisions for the same style of prompt,
especially when the regex layer did not fully determine the outcome.

What I Changed
---------------
1. Removed the unsafe runtime fallback to a random classification head.

   The backend classifier and standalone SDK classifier now only load the
   transformer model when a real trained checkpoint exists on disk.

   A valid checkpoint can contain either:
   - pytorch_model.bin
   - model.safetensors

   If no trained checkpoint is found, the runtime no longer creates a base
   DeBERTa sequence-classification model with random head weights.

2. Added a deterministic heuristic fallback.

   When a fine-tuned model is unavailable, the classifier now uses stable,
   deterministic rules based on the existing regex filter plus extra common
   prompt-injection patterns.

   This fallback classifies prompts as:
   - malicious for high-risk injection or jailbreak attempts
   - suspicious for lower-risk prompt disclosure or manipulation language
   - benign for normal prompts

   This is not a replacement for a fine-tuned semantic classifier, but it is
   much safer than returning predictions from random model weights.

3. Preserved fine-tuned model support.

   If a trained model is placed in the configured classifier directory, the
   Guard still loads and uses the transformer classifier as intended.

   The training path still supports loading the base DeBERTa model explicitly
   when training is requested. The change only prevents random base-model
   inference during normal Guard runtime.

4. Updated model detection.

   The backend config and SDK model detection now recognize both common
   HuggingFace weight formats:
   - pytorch_model.bin
   - model.safetensors

5. Added regression tests.

   New tests verify that when the fine-tuned model is missing:
   - the classifier uses the deterministic fallback
   - the random DeBERTa head path is not called
   - known injection prompts produce stable malicious/suspicious results
   - benign prompts remain benign

6. Updated documentation.

   The docs previously stated that using the base DeBERTa model with random
   classification head weights was the default behavior. That was the source of
   the bug, so the docs now explain the corrected behavior:

   - trained checkpoint available: use fine-tuned DeBERTa classifier
   - trained checkpoint unavailable: use deterministic heuristic fallback
   - fine-tuning is still recommended for stronger semantic coverage

Files Updated
-------------
Backend:
- backend/app/modules/guard/intent_classifier.py
- backend/app/modules/guard/guard_config.py
- backend/app/modules/guard/llm_guard.py
- backend/tests/test_guard.py

Standalone Guard SDK:
- guard-sdk/src/aegisai_guard/intent_classifier.py
- guard-sdk/src/aegisai_guard/llm_guard.py
- guard-sdk/tests/test_guard_sdk.py
- guard-sdk/README.md

Documentation:
- docs/guard-module.md
- docs/getting-started.md
- docs/architecture.md

Why This Fix Solves the Problem
-------------------------------
The original bug happened because the system made predictions using a classifier
head that had never been trained. Random weights are not meaningful for security
decisions, so the Guard could behave unpredictably.

After this fix, Guard decisions are no longer based on random model outputs.
Runtime behavior is now deterministic when no trained model is installed, and
the system still supports loading a proper fine-tuned classifier whenever one is
provided.

This means a fresh install of AegisAI will not silently rely on random ML
predictions for prompt injection detection.

Verification
------------
A syntax-level verification was completed successfully with:

python3 -m py_compile backend/app/modules/guard/intent_classifier.py backend/app/modules/guard/guard_config.py guard-sdk/src/aegisai_guard/intent_classifier.py backend/tests/test_guard.py guard-sdk/tests/test_guard_sdk.py

Result
------
The reported issue is addressed:

- The Guard classifier no longer uses random classification head weights during
  normal runtime.
- Missing fine-tuned models now trigger deterministic fallback logic.
- Documentation now matches the corrected behavior.
- Regression tests were added to prevent the random-head fallback from returning.

And I would request you to label this as GSSoC Approved

### DIFF
--- a/backend/app/modules/guard/guard_config.py
+++ b/backend/app/modules/guard/guard_config.py
@@ -54,13 +54,21 @@ MODEL_METADATA_PATH = MODELS_DIR / "config.json"
 TRAINING_METRICS_PATH = MODELS_DIR / "training_metrics.json"
 
 
+def _has_model_weights(path: str) -> bool:
+    return any(
+        os.path.exists(os.path.join(path, filename))
+        for filename in ("pytorch_model.bin", "model.safetensors")
+    )
+
+
 def get_trained_model_path() -> str:
     """
     Detect trained model location. Checks multiple paths:
     1. Environment variable CLASSIFIER_MODEL_PATH
     2. Local models directory (guard/models/classifier)
     3. Current directory (intent_classifier)
-    Returns default path if not found (will use pre-trained model as fallback).
+    Returns default path if not found (the classifier will use deterministic
+    heuristic fallback until a fine-tuned model is available).
     """
     if os.path.exists(CLASSIFIER_MODEL_PATH):
         return CLASSIFIER_MODEL_PATH
@@ -73,10 +81,8 @@ def get_trained_model_path() -> str:
     ]
 
     for path in alt_paths:
-        if os.path.exists(path) and os.path.exists(
-            os.path.join(path, "pytorch_model.bin")
-        ):
+        if os.path.exists(path) and _has_model_weights(path):
             return path
 
-    # Return default path (will use pre-trained if fine-tuned not found)
+    # Return default path (classifier will use deterministic fallback if missing)
     return CLASSIFIER_MODEL_PATH
--- a/backend/app/modules/guard/intent_classifier.py
+++ b/backend/app/modules/guard/intent_classifier.py
@@ -2,6 +2,7 @@
 
 import os
 import json
+import re
 from typing import List, Tuple, Dict, Optional
 from dataclasses import dataclass
 import numpy as np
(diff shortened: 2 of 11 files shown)


SdSarthak (commented):
This fixes a real safety issue — deterministic heuristics when no fine-tuned model is present is far better than random DeBERTa weights silently producing garbage classifications. The lazy-import __init__.py also resolves the module-level torch import problem correctly.

The PR has a merge conflict — likely due to recent changes in the guard module. Please rebase onto main and this should be ready.

MaheshReddy-ML:
@SdSarthak Can you please label this as **GSSoC Approved**

PR #590, opened 2026-05-24
feat: Add Jira and Linear integration for high-risk AI systems

### LINKED ISSUE #89: Add Jira and Linear integration — auto-create tasks for high-risk AI systems
When a system is classified HIGH or UNACCEPTABLE, optionally create a task in the user's Jira or Linear workspace. Add integration settings per user:
- `POST /api/v1/integrations/jira`
- `POST /api/v1/integrations/linear`

Each compliance gap (from the gap analysis endpoint) should optionally map to a separate ticket.

SiddharthRiot:
Closes #89

## What's changed

- Added `UserIntegration` model to store per-user Jira/Linear credentials
- Added `POST /api/v1/integrations/jira` save Jira integration settings
- Added `POST /api/v1/integrations/linear` save Linear integration settings
- Added `GET /api/v1/integrations/`  lst user's integrations
- Auto-creates tickets in Jira/Linear when a system is classified HIGH or UNACCEPTABLE
- Each compliance gap maps to a separate ticket

## Security
- API tokens encrypted at rest using Fernet (AES-128) derived from `SECRET_KEY`
- `api_token` excluded from all API responses

## Bug fixes (pre-existing)
- Fixed `langchain.text_splitter` import error (moved to `langchain_text_splitters`)
- Fixed sync SQLAlchemy session in async context

## Files added
- `backend/app/models/integration.py`
- `backend/app/schemas/integration.py`
- `backend/app/modules/integrations/jira_client.py`
- `backend/app/modules/integrations/linear_client.py`
- `backend/app/api/v1/integrations.py`
- `backend/app/core/encryption.py`

## Files modified
- `backend/app/models/__init__.py`
- `backend/app/models/user.py`
- `backend/app/api/v1/__init__.py`
- `backend/app/api/v1/classification.py`
- `backend/app/modules/rag/document_loader.py`

## Testing
- `POST /api/v1/integrations/jira` ✅
- `POST /api/v1/integrations/linear` ✅
- `GET /api/v1/integrations/` ✅
- Token not exposed in response ✅

### DIFF
--- a/backend/alembic/versions/137b8187e377_add_user_integrations_table.py
+++ b/backend/alembic/versions/137b8187e377_add_user_integrations_table.py
@@ -19,7 +19,7 @@ depends_on: Union[str, Sequence[str], None] = None
 
 
 def upgrade() -> None:
-    integration_type = postgresql.ENUM('jira', 'linear', name='integrationtype', create_type=False)
+    integration_type = postgresql.ENUM('jira', 'linear', name='integrationtype', create_type=True)
     op.create_table(
         'user_integrations',
         sa.Column('id', sa.Integer(), nullable=False),
@@ -27,7 +27,7 @@ def upgrade() -> None:
         sa.Column('integration_type', integration_type, nullable=False),
         sa.Column('base_url', sa.String(512), nullable=True),
         sa.Column('email', sa.String(255), nullable=True),
-        sa.Column('api_token', sa.String(512), nullable=False),
+        sa.Column('api_token', sa.Text(), nullable=False),
         sa.Column('project_key', sa.String(255), nullable=True),
         sa.Column('created_at', sa.DateTime(), nullable=True),
         sa.Column('updated_at', sa.DateTime(), nullable=True),
@@ -39,3 +39,4 @@ def upgrade() -> None:
 def downgrade() -> None:
     op.drop_index(op.f('ix_user_integrations_id'), table_name='user_integrations')
     op.drop_table('user_integrations')
+    op.execute("DROP TYPE IF EXISTS integrationtype")
--- a/backend/alembic/versions/137b8187e377_add_user_integrations_table.py
+++ b/backend/alembic/versions/137b8187e377_add_user_integrations_table.py
@@ -1,0 +1,41 @@
+"""add_user_integrations_table
+
+Revision ID: 137b8187e377
+Revises: 55a49e4b7bc8
+Create Date: 2026-05-24 19:23:05.102279
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision: str = '137b8187e377'
+down_revision: Union[str, None] = '55a49e4b7bc8'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    integration_type = postgresql.ENUM('jira', 'linear', name='integrationtype', create_type=False)
+    op.create_table(
+        'user_integrations',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('user_id', sa.Integer(), sa.ForeignKey('users.id'), nullable=False),
+        sa.Column('integration_type', integration_type, nullable=False),
+        sa.Column('base_url', sa.String(512), nullable=True),
+        sa.Column('email', sa.String(255), nullable=True),
+        sa.Column('api_token', sa.String(512), nullable=False),
+        sa.Column('project_key', sa.String(255), nullable=True),
+        sa.Column('created_at', sa.DateTime(), nullable=True),
+        sa.Column('updated_at', sa.DateTime(), nullable=True),
+        sa.PrimaryKeyConstraint('id'),
+    )
+    op.create_index(op.f('ix_user_integrations_id'), 'user_integrations', ['id'], unique=False)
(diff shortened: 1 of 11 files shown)


SiddharthRiot:
Backend test failures are pre-existing on `main` branch `test_rag_ingest` auth mock issue and `test_document_generation` import error are unrelated to this PR. Frontend Lint passes.

SdSarthak (changes requested):
The encryption approach (Fernet via Sha256-derived key from SECRET_KEY) is a significant improvement over the plaintext approach in the previous attempt. The integration model, schema, and router are all well-structured. A few things to fix:

**1. No Alembic migration (must fix)**
The `user_integrations` table won't be created without a migration. Run `alembic revision --autogenerate -m "add_user_integrations_table"` after applying existing migrations, and verify the generated file only creates the new table.

**2. `asyncio.run()` inside a sync route (must fix)**
`asyncio.run(client.create_issue(...))` inside the classification endpoint will raise `RuntimeError: This event loop is already running` when FastAPI runs the route in an async context. Use `BackgroundTasks` to fire the integration call asynchronously:
```python
background_tasks.add_task(fire_integration_task, ...)
```
Or make the route `async def` and `await` the client call directly.

**3. `err: unknown` in RagChat.tsx breaks the error handler (must fix)**
```ts
} catch (err: unknown) {
  if (err.response?.status === 503) {  // TypeScript error: err is 'unknown'
```
Change back to `err: any` or narrow with `axios.isAxiosError(err)`.

SiddharthRiot:
fixed all the issues you pointed out migration for `user_integrations` is in, swapped `asyncio.run()` for `BackgroundTasks`, and sorted the `RagChat` lint error should be good to go now!

SiddharthRiot:
backend test failure is pre-existing `test_generate_technical_documentation` expects 201 but the endpoint correctly returns 200 this exists on main branch before my changes

SdSarthak (changes requested):
Two fixes needed: (1) The `api_token` column is `String(512)` but Fernet ciphertext can exceed that for longer tokens — use `Text` instead. (2) The Alembic migration uses `create_type=False` for the `IntegrationType` enum without confirming the type pre-exists in Postgres — verify this or add `CREATE TYPE` to the migration. No frontend UI is included to configure integrations, which is fine for a backend-only first pass, but please note it in the PR description.

SiddharthRiot:
`isApiError` function and `ApiError` interface are defined but unused, causing ESLint `no-unused-vars` CI failure. Should I remove them and use the current `err as {...}` type assertion approach, or refactor the catch block to use `isApiError` for proper type narrowing?

SdSarthak:
The two code issues from the previous review are fixed — `api_token` column is now `Text` and the `IntegrationType` enum migration uses `create_type=True`. The implementation looks good.

Please rebase on the latest `main` to clear the current conflicts. After the rebase, also verify the Alembic migration's `down_revision` matches the current head revision in the repo — it will need updating after recent merged migrations.

SiddharthRiot:
@SdSarthak, rebased on latest main and updated the migration's `down_revision` to `c3d9f1b2a4e6` (current head after recent merges). Ready for re-review!

SdSarthak (changes requested):
Good progress — both previous blockers are resolved: `api_token` uses `Text()` and the `IntegrationType` enum uses `create_type=True`. The migration chain (`a7c004156ad1` → `c3d9f1b2a4e6` → `137b8187e377`) is correct.

One last fix before merge: **missing trailing newlines** — the diff shows 39 instances of `\ No newline at end of file`. This affects `jira_client.py`, `linear_client.py`, `App.tsx`, and others. Most editors handle this automatically. Run `git diff --check` before pushing and ensure all files end with a newline.

After adding trailing newlines and rebasing on main to resolve any remaining conflicts, this is ready to merge.

SiddharthRiot:
@SdSarthak  rebased on main trailing newlines fixed! ready for review

SdSarthak (changes requested):
Good feature direction — Jira and Linear ticket creation for high-risk systems is a valuable workflow. A few issues to fix before merge:

1. **Migration revision chain**: `down_revision = '55a49e4b7bc8'` is stale. The current migration head is `9f2b7c6a1d3e`. Please update `down_revision` to point to the correct head.

2. **PostgreSQL-only ENUM**: The migration uses `postgresql.ENUM` directly, which is pg-specific and will break SQLite in tests. Use `sa.Enum(IntegrationType)` (SQLAlchemy cross-dialect) in the `create_table` call instead.

3. **SSRF risk on `JiraClient.base_url`**: Users can supply any URL as `base_url`, including internal network addresses. Apply the same `_is_private_ip` check (or similar) from the SSRF webhook PR when saving the integration, so internal endpoints cannot be targeted.

4. **`from app.core.encryption import encrypt/decrypt`**: Please confirm this module exists in main. If not, the app will fail to start. If it doesn't exist yet, either add a minimal implementation (e.g. Fernet with a key from settings) or use a simpler approach for the initial PR.

5. **Missing newline at end of file** on several new files — not blocking but worth fixing.

SdSarthak:
The migration still shows `down_revision = '55a49e4b7bc8'` — this needs to be updated to the current migration head: `9f2b7c6a1d3e`. Also this PR is still CONFLICTING — please rebase on `main` after fixing the migration head.

SiddharthRiot:
conflicts resolved, migration head updated to `9f2b7c6a1d3e`, rebased on main

SdSarthak:
Still conflicting with main after the last rebase. Could you fetch origin, rebase onto origin/main again, resolve, and force-push?

PR #592, opened 2026-05-24
fix: send page param for AI systems pagination

saurabhhhcodes:
## Summary\n- send the backend-supported  parameter from the AI Systems page instead of \n- remove the unused frontend  calculation\n- align the API service typing with the backend contract\n\n## Validation\n- git diff --check\n- code path verified against backend  pagination contract\n- note: npm-based frontend lint/build could not be run in this shell because the available Node runtime here does not include an npm CLI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ---
 
+## [Unreleased]
+
+### Added
+- LLM Guard console with copy-to-clipboard exports for scan response payloads and raw audit metrics.
+
+---
+
 ## [0.1.0] — 2026-04-05
 
 ### Added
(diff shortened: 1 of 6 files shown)


saurabhhhcodes:
Refreshed the branch with a small frontend formatting cleanup on top of the pagination fix so CI can rerun against the tidier diff.

saurabhhhcodes:
Updated the branch again. Besides the pagination fix, I also addressed the active frontend lint blocker in  () so the rerun can get past the frontend gate.

saurabhhhcodes:
Updated the branch again. Besides the pagination fix, I also addressed the active frontend lint blocker in RagChat.tsx (`@typescript-eslint/no-explicit-any`) so the rerun can get past the frontend gate.

saurabhhhcodes:
I found that the earlier rerun was still attached to an older head SHA, so I pushed a fresh follow-up commit to move the PR head cleanly and retrigger CI. The pagination fix is unchanged; this push is just to ensure GitHub Actions evaluates the current branch state, including the frontend lint cleanup.

PR #539, opened 2026-05-23
Feature/webhooks api delivery

EHEGUY:
Summary

This PR introduces webhooks functionality to the application, allowing for real-time event notifications and system integration.

Closes #129
Type of Change

    [ ] Bug fix

    [x] New feature

    [ ] Documentation update

    [ ] Refactor

    [x] Tests

    [ ] Infra / CI

Checklist

    [x] I have read CONTRIBUTING.md

    [x] My code follows the project style (PEP 8 for Python, ESLint for TS)

    [ ] I have added/updated tests where relevant

    [ ] pytest backend/tests/ passes locally

    [x] I have not committed .env or any secrets

    [ ] I have updated documentation if needed

PR #613, opened 2026-05-24
fix: implement webhook API delivery

EHEGUY:
## Summary

Closes <-#539-> make it mergable differet PR

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor
- [ ] Tests
- [ ] Infra / CI

## Checklist

- [ ] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [ ] `pytest backend/tests/` passes locally

### DIFF
--- a/backend/app/api/v1/rag.py
+++ b/backend/app/api/v1/rag.py
@@ -12,6 +12,7 @@ TODO for contributors (high difficulty):
 import os
 import shutil
 import tempfile
+import time
 from typing import List, Optional
 
 from fastapi import APIRouter, Depends, File, HTTPException, UploadFile, status
--- a/backend/app/api/v1/webhooks.py
+++ b/backend/app/api/v1/webhooks.py
@@ -4,8 +4,13 @@ Copyright (C) 2024 Sarthak Doshi (github.com/SdSarthak)
 SPDX-License-Identifier: AGPL-3.0-only
 """
 
-from typing import List
+import asyncio
+import hashlib
+import hmac
+import json
+import logging
 
+import httpx
 from fastapi import APIRouter, Depends, HTTPException, status
 from sqlalchemy.orm import Session
 
@@ -13,17 +18,18 @@ from app.core.database import get_db
 from app.core.security import get_current_user
 from app.models.user import User
 from app.models.webhook import WebhookConfig
-from app.schemas.webhook import WebhookCreate, WebhookResponse
+from app.schemas.webhook import WebhookConfigCreate, WebhookConfigOut
 
 router = APIRouter()
+logger = logging.getLogger(__name__)
 
 
-@router.post("", response_model=WebhookResponse, status_code=status.HTTP_201_CREATED)
+@router.post("", response_model=WebhookConfigOut, status_code=status.HTTP_201_CREATED)
 def create_webhook(
-    body: WebhookCreate,
+    body: WebhookConfigCreate,
     current_user: User = Depends(get_current_user),
     db: Session = Depends(get_db),
-):
+) -> WebhookConfig:
     """Register a new webhook endpoint for the current user.
 
     Args:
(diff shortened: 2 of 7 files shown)


SdSarthak (changes requested):
The implementation is solid — HMAC-SHA256 signing via the `X-AegisAI-Signature` header, concurrent delivery with `asyncio.gather`, configurable per-event filtering, and a `BackgroundTasks`-compatible design. The schema, model, and test coverage are all well-structured.\n\nOne issue to fix before merging:\n\n**`asyncio.run()` in tests will fail in async test contexts**\nIn `test_webhooks.py`:\n```python\nasyncio.run(deliver_webhook(db_session, test_user.id, event, payload))\n```\n`asyncio.run()` raises `RuntimeError: This event loop is already running` when called from within an existing event loop (which pytest-asyncio or the test runner may be running). Use `pytest.mark.asyncio` and `await deliver_webhook(...)` instead, or use `asyncio.get_event_loop().run_until_complete()` as a fallback — though the cleanest approach is just marking the test as async:\n```python\n@pytest.mark.asyncio\nasync def test_webhook_delivery(...):\n    await deliver_webhook(...)\n```\n\nPlease also add `pytest-asyncio` to `requirements-test.txt` or `pyproject.toml` if it's not already there.\n\nEverything else looks good — fix the async test pattern and this is ready to merge.

SdSarthak (changes requested):
Three things to address: (1) The PR deletes `app/schemas/webhook.py` content and introduces a new `app/schemas/webhook_config.py` with near-identical models — please clean up the old file or consolidate rather than splitting schemas. (2) The DELETE endpoint changes from `204 No Content` to `200 OK` with a body — this is a breaking API change that needs justification. (3) PRs #706 and #665 implement the same `deliver_webhook` function — please confirm those will be closed if this one lands.

EHEGUY:
@SdSarthak — Addressed all three review items plus fixed two pre-existing CI failures. Here's the breakdown:

---

## Review Feedback Addressed

### 1. Schema consolidation (resolved)

Deleted `app/schemas/webhook_config.py` entirely. The models `WebhookConfigCreate` and `WebhookConfigOut` now live in `app/schemas/webhook.py` alongside the existing `WebhookCreate` / `WebhookResponse` models. The import in `app/api/v1/webhooks.py` was updated to `from app.schemas.webhook import WebhookConfigCreate, WebhookConfigOut`. No duplicate schema files remain.

### 2. DELETE endpoint reverted to 204 No Content (resolved)

The `DELETE /api/v1/webhooks/{webhook_id}` endpoint now returns `HTTP 204 No Content` with no response body, matching the original API contract. The test assertion was updated accordingly (`assert response.status_code == 204` and `assert response.content == b""`).

### 3. Overlap with PRs #706 and #665

Both #706 and #665 implement a similar `deliver_webhook` function. If this PR lands first, those should be closed as superseded — the implementation here covers the same functionality (HMAC-SHA256 signing, async fan-out delivery, graceful failure handling). I'll leave a note on those PRs once this merges.

---

## CI Fixes (pre-existing failures unrelated to this PR)

### Fix: `test_query_feedback_and_low_quality_flow` — was returning 503

**Root causes (3 issues compounding):**
1. `import time` was missing in `app/api/v1/rag.py` — the endpoint uses `time.monotonic()` for latency measurement but never imported the module. The resulting `NameError` was caught by the broad `except Exception` handler and surfaced as a 503.
2. The test's `DummyDoc` class was missing a `page_content` attribute — the endpoint accesses `doc.page_content` for the groundedness check, causing `AttributeError` → 503.
3. The test used fragile `sys.modules` patching that could race with Python's import cache. Rewrote to use `unittest.mock.patch` targeting the actual module attributes, which is the standard pattern for mocking `from X import Y` inside function bodies.

### Fix: `test_per_user_rate_limit_blocks_61st_guard_scan_request` — early 429

**Root cause:** The guard scan rate limiter uses an in-memory `defaultdict(deque)` at module level (`_scan_attempts_by_user`). When other tests hit `/guard/scan` earlier in the session, the deque already has entries, so this test's 60-request budget was partially consumed before it started. Added `_scan_attempts_by_user.clear()` at the top of the test to ensure a clean slate.

---

## Files Changed

| File | Change |
|------|--------|
| `backend/app/api/v1/webhooks.py` | Added `deliver_webhook` async function, reverted DELETE to 204, updated schema import |
| `backend/app/schemas/webhook.py` | Added `WebhookConfigCreate` and `WebhookConfigOut` (consolidated from deleted file) |
| `backend/app/api/v1/rag.py` | Added missing `import time` |
| `backend/tests/api/v1/test_webhooks.py` | New test file — covers create, list, delete, scoping, delivery, and failure handling |
| `backend/tests/integration/test_rag_feedback.py` | Fixed DummyDoc, rewrote mocking to use `unittest.mock.patch` |
| `backend/tests/integration/test_rate_limiting.py` | Clear rate limiter state before test |
| `backend/requirements.txt` | Added `httpx` and `respx` (test deps for async webhook delivery tests) |

---

Rebased on current `main` — no merge conflicts. Let me know if anything else needs adjustment.

EHEGUY:
@SdSarthak lmk if there is anything wrong

SdSarthak:
Closing as superseded: the async webhook delivery utility was merged via #706 using the existing `WebhookCreate`/`WebhookResponse` schema names. This PR references `WebhookConfigCreate`/`WebhookConfigOut` which do not exist in the current schema, so it would fail at runtime.

EHEGUY:
Got it, thanks for the heads up! Glad to see it was handled in #706.

PR #624, opened 2026-05-24
feat: implement static onboarding wizard layout

AnmolM-777:
## Summary

Closes #<112>


SdSarthak:
Good progress on the step state and UI — the sector/use-case selectors and step logic look clean. However the PR's own acceptance criteria note *'Submission logic prints form state to console.log instead of performing active API calls'*, and the `console.log` is still present in the final step handler. Production builds must not log form data to the console. Please remove the `console.log` and wire up the actual API call (`aiSystemsApi.create`), or at minimum remove the log and navigate forward without it. Also make sure `/onboarding` is inside `PrivateRoute` in App.tsx.

PR #626, opened 2026-05-24
feat: implement risk classification explainer API and Python 3.8 type annotation patches

AnmolM-777:
Closes #<456>

### DIFF
--- a/backend/app/api/v1/classification.py
+++ b/backend/app/api/v1/classification.py
@@ -13,6 +13,8 @@ from app.schemas.ai_system import (
     RiskClassificationResponse,
     QuestionnaireRiskFactor,
 )
+from app.schemas.classification import ExplainRequest, ExplainResponse
+from app.modules.explainer import explain_system_risk
 
 router = APIRouter()
 
@@ -468,5 +470,17 @@ def bulk_classify_systems(
 
     db.commit()
     return BulkClassificationResponse(results=results)
+
+
+@router.post("/explain", response_model=ExplainResponse)
+def explain_classification(
+    request: ExplainRequest,
+    current_user: User = Depends(get_current_user)
+):
+    """
+    Explain the risk classification of an AI system description under the EU AI Act.
+    Returns predicted risk level, reasoning, legal articles, and action recommendations.
+    """
+    return explain_system_risk(request.description)
 
 
--- a/backend/app/api/v1/guard.py
+++ b/backend/app/api/v1/guard.py
@@ -8,9 +8,10 @@ TODO for contributors (medium difficulty):
   - Persist scan results to the database for audit logs (Completed)
   - Add a GET /guard/stats endpoint returning block/allow/sanitize counts (Completed)
 """
-
+from __future__ import annotations
 import hashlib
 from collections import Counter, defaultdict, deque
+
 from datetime import datetime, timedelta, timezone
 from threading import Lock
 from typing import Optional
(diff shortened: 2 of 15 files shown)


SdSarthak (changes requested):
The rule-based explainer is a well-thought-out addition — keyword-to-risk-factor mapping, article citations, and plain-English recommendations all make sense for a compliance tool. The no-external-LLM approach is the right call.\n\nOne issue that needs to be addressed before merging:\n\n**`from __future__ import annotations` added to existing files (must remove)**\nThis PR adds `from __future__ import annotations` to several existing files: `guard.py`, `notifications.py`, `rag.py`, `badge_generator.py`, and others. This flag makes Python defer all type annotations to strings at runtime, which breaks Pydantic v2's model validation (Pydantic v2 evaluates annotations eagerly to build field validators). Adding this import to files with Pydantic models will silently cause validation to fail or produce wrong types.\n\nPlease remove `from __future__ import annotations` from all existing files — it should not appear in any file that defines or uses Pydantic models. The new `explainer/core.py` and `explainer/engine.py` files don't define Pydantic models, so they can keep it if needed for Python 3.8 compat, but verify they don't transitively import Pydantic models that would be affected.\n\nEverything else looks correct. Remove the future-annotations imports from existing files and this is ready to merge.

SdSarthak (approved):
**MERGE**

Good addition with a clean implementation. Key findings:

- `app/modules/explainer/core.py` is a pure rule-based keyword classifier — no ML inference, fully deterministic, fast. Appropriate for a v1 explainer that needs to run without LLM dependencies.
- The four risk tiers (UNACCEPTABLE, HIGH, LIMITED, MINIMAL) correctly map to EU AI Act Article 5, Annex III, Article 52, and Article 95 respectively. Legal article references appear accurate.
- `confidence` values are hardcoded heuristics (0.99 for UNACCEPTABLE, 0.75 for MINIMAL) — this is fine for a keyword classifier as long as it is communicated to users. The `ExplainResponse` schema does expose `confidence` which is good.
- Test coverage is thorough: 5 unit tests covering each risk tier plus the empty-string fallback, plus 2 integration tests covering the endpoint and 422 validation. All test assertions check structure, not just status codes.
- The `from __future__ import annotations` patches across `guard.py`, `notifications.py`, `rag.py`, `badge_generator.py`, and schemas are correct Python 3.8 compatibility fixes. The PR title accurately describes this as a secondary concern.
- The `ExplainRequest`/`ExplainResponse` schemas are properly wired into the `__init__.py` exports.

Minor nits (non-blocking):
- `core.py` uses `Dict`, `List`, `Any` from `typing` rather than the built-in generics (`dict`, `list`). Since the `from __future__ import annotations` patches elsewhere in the PR are specifically for 3.8 compat, this module should also use `from __future__ import annotations` + built-ins, or consistently use `typing`. Either is fine, just inconsistent.
- The `HIGH` confidence cap formula `min(0.98, 0.90 + 0.01 * len(high_matches))` could in theory equal 0.98 for 8+ keyword matches, which is misleadingly close to the UNACCEPTABLE confidence. Document or cap lower.
- PR body references `Closes #<456>` (literal placeholder) — fix the issue reference.

None of these block merging. The explainer is scoped appropriately for an open-source GRC tool and the test suite is production-quality.

SdSarthak:
This PR has merge conflicts with `main`. Please rebase on the latest `main` branch and resolve the conflicts — once done it will be merged.

SdSarthak:
This PR is approved and ready to merge — it just needs a rebase on the latest `main` to resolve the current conflicts. Please rebase and it will be merged right away.

SdSarthak:
This PR is approved and the implementation looks solid — just needs a rebase on the latest `main` to clear the conflicts. Please rebase and it will be merged immediately.

SdSarthak:
This was previously approved and is a solid risk classification explainer — please rebase on main as it's now conflicting after recent merges to classification.py. Once rebased it's ready to merge.

SdSarthak (commented):

PR #629, opened 2026-05-24
feat: replace create_all with Alembic migrations

### LINKED ISSUE #70: Add Alembic database migrations
Replace `Base.metadata.create_all()` on startup with proper Alembic migrations:
1. Initialise with `alembic init alembic`
2. Generate the initial migration from current models
3. Document the migration workflow in `CONTRIBUTING.md`

Alembic is already in `requirements.txt`.

Anshika0998:
## Summary

This PR replaces automatic database table creation on startup with a proper Alembic migration workflow.

### Changes made

* Added Alembic initial migration setup
* Removed `Base.metadata.create_all()` from application startup
* Updated startup logging to indicate migrations should be applied before startup
* Added migration workflow documentation to `CONTRIBUTING.md`

### Migration workflow

* Create migration:
  `alembic revision --autogenerate -m "message"`
* Apply migrations:
  `alembic upgrade head`
* Rollback:
  `alembic downgrade -1`

### Verification

* Verified PostgreSQL connection locally
* Successfully ran:
  `alembic upgrade head`
* Verified backend startup with:
  `uvicorn app.main:app --reload`

Closes #70


SdSarthak:
This adds a new 'initial schema' migration with `down_revision = None` — a second root in the Alembic migration chain. The project already has an existing migration chain starting from the original schema. Adding a second root creates a branch that Alembic cannot resolve and `alembic upgrade head` would fail. Additionally, the `main.py` change removes `Base.metadata.create_all()` without replacing it with `alembic upgrade head` in the startup flow, which means migrations won't run on startup. Closing as it would break the migration system.

PR #630, opened 2026-05-24
feat(auth): add API key management endpoints

asthasinghcs:
## Summary

Implemented API key management endpoints for authenticated users.

## Changes

* Added `ApiKey` SQLAlchemy model
* Added API key request/response schemas
* Added API key model registration in `models/__init__.py`
* Added endpoints:

  * `POST /api/v1/auth/api-keys`
  * `GET /api/v1/auth/api-keys`
  * `DELETE /api/v1/auth/api-keys/{api_key_id}`
* API keys are:

  * generated securely using `secrets.token_urlsafe`
  * stored as SHA-256 hashes in the database
  * returned only once during creation
  * revocable without deleting records

## Testing

* `python -m compileall backend/app` passed successfully


SdSarthak (changes requested):
The API key management design is solid — storing only the SHA-256 hash of the raw key (never the key itself), generating with `secrets.token_urlsafe`, returning the plaintext key only once at creation, and scoping revoke/list to the authenticated user. The stats endpoint refactor is also an improvement (uses `COUNT` queries instead of fetching all rows).\n\nOne thing to fix before merging:\n\n**Missing Alembic migration (must fix)**\nThe `ApiKey` model is added but there's no migration to create the `api_keys` table. Without it, the endpoint will crash on the first request with `Table 'api_keys' doesn't exist`. Run:\n```\ncd backend && alembic revision --autogenerate -m "add_api_keys_table"\n```\nAfter applying the current head migration (`a7c004156ad1`). Verify the generated migration only creates `api_keys` and doesn't regenerate existing tables.\n\nAlso check that `ApiKey` is added to `app/models/__init__.py` so Alembic detects it during autogenerate — it's added, which is correct.\n\nOnce the migration is in place this is ready to merge.

SdSarthak (changes requested):
**REQUEST_CHANGES**

Solid, well-structured implementation. The model, schema, and endpoints are all clean. Issues to fix before merge:

1. **No Alembic migration** — `ApiKey` model is defined and registered in `__init__.py` but there is no migration file in the diff. The table will not be created in existing deployments. Add `alembic revision --autogenerate -m "add_api_keys_table"` and include the generated file.

2. **No authentication path using API keys** — the PR adds CRUD for managing keys but never wires them into the auth middleware. `get_current_user` in `app/core/security.py` still only handles JWT Bearer tokens. Storing key hashes is pointless without a code path that validates them. Either add the auth middleware in this PR or document it clearly as a follow-up issue with a linked issue number.

3. **`revoked_at` is never set** — the `revoke_api_key` endpoint sets `api_key.revoked = True` but does not populate `revoked_at = datetime.utcnow()`. Fix this.

4. **`ApiKeyResponse` does not expose `revoked_at`** — since the field exists on the model, include it in the response schema so consumers know when revocation happened.

5. **Missing newline at end of `auth.py`, `api_key.py`, and `api_key.py` schema** — all three files end without a trailing newline.

6. **`get_current_user_stats` refactor is bundled in** — the stats endpoint was refactored (count queries, enum `.value` handling) in this PR alongside the API key feature. These are separate concerns; ideally split into a separate PR, though the changes themselves are correct improvements.

7. **No tests** — add tests for create, list, and revoke endpoints, and for the duplicate-name edge case.

The core model and endpoint design is good — fix the migration, the `revoked_at` bug, and add tests.

SdSarthak (changes requested):
Four fixes needed: (1) No Alembic migration — existing deployments will not get the `api_keys` table. (2) `revoked_at` is never set on the revoke endpoint — the column always stays NULL. (3) The API key auth is not wired into `get_current_user`, making the stored hashes currently unused. (4) No tests. The stats endpoint refactor bundled in is correct but should be a separate PR.

SdSarthak (changes requested):
Two critical issues before this can merge:

1. **Accidental credential file**: `backend/Untitled` is included in the diff and contains `DATABASE_URL` and `SECRET_KEY` values. This file must not be committed to the repository. Please delete it and ensure it is added to `.gitignore`.

2. **Destructive migration change**: The existing migration `eb8060353ac6_extend_guard_scan_log_metadata.py` has its entire `upgrade()` body replaced with `pass`. This migration adds the `detection_type`, `regex_flag`, `scanned_at`, and other guard scan metadata columns that the codebase relies on. Emptying it would break any database that hasn't run it yet. Please restore the original `upgrade()`/`downgrade()` bodies in that migration — only the new `api_keys` migration should be added.

The `api_keys` table schema and Alembic migration (with correct `down_revision = 'eb8060353ac6'`) are otherwise well-structured.

asthasinghcs:
Thanks for the detailed review. Apologies for the delay while resolving the migration/auth conflicts and CI issues.

I’ll remove the accidental credential file and restore the original guard scan metadata migration bodies, then push the fixes shortly.

asthasinghcs:
Thanks for the review and patience while I worked through the merge conflicts and CI issues.

I’ve now:
- removed the accidental credential file and added it to `.gitignore`
- restored the original guard scan migration bodies
- fixed the auth/security regressions and indentation issues

All CI checks are now passing.

asthasinghcs:
Hi @SdSarthak,

I've addressed the previously requested changes, updated the PR, and all checks are now passing.

The PR still shows "1 change requested", but I'm unable to find any remaining unresolved review comments or new requested changes. Could you please let me know if there's anything else that still needs to be updated, or take another look when you have a chance?

Thank you.

SdSarthak (commented):

asthasinghcs:
Hi @SdSarthak,

I've rebased the branch onto the latest main branch, resolved the merge conflicts, addressed the review feedback, and pushed the updates.

The branch now merges cleanly with the base branch. I also restored the RAG router registration after the rebase, which removed a large number of CI failures.

The remaining CI failures appear to be limited to a few tests, and from my investigation they don't seem directly related to the API key changes in this PR. Could you please take a look when you have a chance and let me know if you'd like any further changes from my side?

Thank you.

SdSarthak:
This branch conflicts with main right now. Can you fetch origin, rebase onto origin/main, resolve the conflicts, and force-push?

PR #642, opened 2026-05-25
Updated Fix for issue 441 

### LINKED ISSUE #441: Add audit logging for Guard scan decisions

Anshuman-cs50:
## Description
This PR resolves the migration schema collision issue introduced in the initial implementation of the Guard scan audit logging feature.

The previous auto-generated migration incorrectly included DDL for all existing database tables (such as `users`, `ai_systems`, `notifications`, etc.), which resulted in `table already exists` errors when applied to instances with an already established database schema. 

## Changes Made
* **Corrected Migration Scope:** Re-generated the Alembic migration file to strictly target only the new `guard_scan_logs` table.
* **Preserved Database Schema:** Ensured the generated operations only map to DDL operations for `guard_scan_logs` and its indexes without attempting to recreate existing relational structures.

## Related Issues
* Fixes #441 

## Testing Performed
* Generated a mock database with all existing schema structures.
* Validated that running `alembic upgrade head` accurately registers the `guard_scan_logs` table incrementally without interfering with previously migrated entities.


SdSarthak:
The `guard_scan_logs` table was already created by an earlier migration and has since been extended with additional columns (detection_type, regex_flag, ml_confidence, combined_score, etc.). Applying this migration would conflict with the existing table.

Additionally, this migration's `down_revision` points to `55a49e4b7bc8`, but the current head of the migration chain is `a7c004156ad1`. New migrations must chain from that revision.

Closing as the underlying table is already in the correct state.

Issue #515, opened 2026-05-22
Bug: train_test_split() called without random_state — results are non-reproducible


Daksh7785:
## Bug Report

**Description**
Multiple projects call `train_test_split()` without setting `random_state`, meaning the
train/test split changes on every run. This produces different model accuracy each run,
making results impossible to reproduce or compare — especially bad in a learning repository
where students expect consistent output.

## Affected Files

| File | Line |
|------|------|
| `Bitcoin Price Prediction Web App/app.py` | 50 |
| `Chicken_Disease_Classification/research/mlflow_exp/example.py` | 47 |
| `Social Media Fake Accounts Detection/jserver/model.py` | 76 |
| `Ensemble Methds in ML/voting.py`, `blending.py`, `stacking.py` | various |

## Code (Buggy)
```python
xtrain, xtest, ytrain, ytest = train_test_split(X, Y, test_size=0.2)
```

## Fix
```python
xtrain, xtest, ytrain, ytest = train_test_split(X, Y, test_size=0.2, random_state=42)
```

## Severity
Medium — non-reproducible experiments; misleads beginners learning ML

snowlindhanusha:
Hi, I am a GSSoC’26 contributor. I would like to work on this issue. Please assign it to me.

akhila-sraj:
Hi @Daksh7785 , I am a GSSoC 2026 contributor and would like to work on this issue.

I have experience with cybersecurity concepts and secure application development, and I’d love to contribute. Please assign this to me.

TejeeshRagnarok:
Hi @Daksh7785 and maintainers, I’d like to work on this issue as part of GSSoC'26.

I have previous backend/AI contribution experience in AegisAI (#400 merged PR), and I’d like to improve the ML pipeline reproducibility by fixing the missing `random_state` usage in `train_test_split()` calls and validating deterministic behavior across runs.

madsysharma:
Hi @SdSarthak , I would like to work on this as part of GSSoC 2026.

YerraguntaAjayKumar:
Hi @SdSarthak and @Daksh7785, I am a GSSoC '26 contributor and would love to resolve this issue to ensure ML pipeline reproducibility across the repository!

Here is my targeted implementation plan to fix the missing seeds:
[...]

PR #649, opened 2026-05-25
test(training): cover deterministic train split

akhila-sraj:
**Description**
This PR adds a regression test to make the training split reproducibility guarantee explicit. The guard training helper already passes `random_state` through to `train_test_split`, but there was no test protecting that behavior.

What changed:
- Added test_training_split.py.
- Verified `train_validation_split(..., random_state=42)` returns the same train/validation split across repeated calls.
- Verified changing the seed produces a different split.

Why it matters:
- Protects against accidental removal of `random_state` in future changes.
- Keeps ML experiments reproducible for learners and contributors.
- Makes the deterministic split behavior explicit in tests.

Validation:
- `pytest test_training_split.py -q` passed in the project venv.

Fixes issue #515


akhila-sraj:
Hi @SdSarthak I have fixed issue#515 could you please review it and merge request

SdSarthak (changes requested):
The test logic itself is correct — deterministic split verification is a good test to have. Two things need to change:

1. **Move to `backend/tests/`** — the file is at `tests/test_training_split.py` (repo root). All backend tests live in `backend/tests/` and are run via `pytest backend/tests/`. Please move the file there.

2. **Remove `from __future__ import annotations`** — this import breaks Pydantic v2 runtime type evaluation across the project. It's a known issue and must not be added to any file in this codebase.

Please make those two changes and push an update.

akhila-sraj:
hi @SdSarthak I have made the changes could you check

SdSarthak (changes requested):
**REQUEST_CHANGES**

The test idea is correct and the behavior being protected (deterministic `random_state` seeding) is worth pinning. However several issues block merging as-is:

1. **The target module `app.modules.guard.training.data.split` does not exist in main** — the guard module on main contains `app/modules/guard/train.py` (a single file), not a `training/data/split.py` subpackage. Both test files import `train_validation_split` from a path that does not exist, so both tests will fail with `ModuleNotFoundError` on CI.

2. **Duplicate test file** — the PR adds the same tests in two locations: `backend/tests/test_training_split.py` and `tests/test_training_split.py`. The root-level `tests/` is for standalone integration tests; backend unit tests belong only in `backend/tests/`. The root-level file also uses a fragile `importlib.util.spec_from_file_location` path hack to load the module, which will break if the directory structure changes. Keep one copy in `backend/tests/` using a normal import.

3. **Second test `test_train_validation_split_changes_with_different_seed` uses `or` logic** — the assertion `assert not first_train.equals(second_train) or not first_val.equals(second_val)` passes even if only one of the two splits differs, which is weaker than intended. Use `and`.

4. **Missing newline at end of both test files**.

Please work with the maintainer to confirm the correct module path (likely `app.modules.guard.train` and the function name within it), remove the duplicate root-level test file, fix the assertion, and resubmit.

SdSarthak (changes requested):
The import path `app.modules.guard.training.data.split` does not exist in main — the guard module uses a single `train.py` file. Both test files will fail with `ModuleNotFoundError`. Please confirm the correct import path from the actual file structure and submit a single test file in `backend/tests/`.

akhila-sraj:
@SdSarthak Can you review it once?

PR #650, opened 2026-05-25
fix(audit): serialize audit log values safely

akhila-sraj:
**PR Title**
`test(training): cover deterministic train split`

**Description**
This PR adds a regression test to make the training split reproducibility guarantee explicit. The guard training helper already passes `random_state` through to `train_test_split`, but there was no test protecting that behavior.

What changed:
- Added test_training_split.py.
- Verified `train_validation_split(..., random_state=42)` returns the same train/validation split across repeated calls.
- Verified changing the seed produces a different split.

Why it matters:
- Protects against accidental removal of `random_state` in future changes.
- Keeps ML experiments reproducible for learners and contributors.
- Makes the deterministic split behavior explicit in tests.

Validation:
- `pytest test_training_split.py -q` passed in the project venv.

Fixes issue #515

### DIFF
--- a/backend/app/models/audit_log.py
+++ b/backend/app/models/audit_log.py
@@ -5,6 +5,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 """
 
 from datetime import datetime
+import enum
 
 from sqlalchemy import Column, Integer, DateTime, ForeignKey, JSON, event
 from sqlalchemy.orm import relationship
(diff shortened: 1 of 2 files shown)


SdSarthak:
This is a duplicate of the PR already merged for the audit log JSON serialization fix. The diff is identical. Closing this copy — the fix is live.

PR #682, opened 2026-05-26
chore: add Docker ignore files

saurabhhhcodes:
## Summary
- add a root .dockerignore for repository-level builds
- add backend and frontend .dockerignore files for the actual compose build contexts
- keep env files, caches, logs, node_modules, test artifacts, and VCS metadata out of Docker build contexts

## Validation
- git diff --check


saurabhhhcodes:
I traced the red CI here and found the failures are shared base/test-contract issues rather than this Docker ignore change.

I opened #690 to address the common failures reported by this run:
- document generation status should be 201
- guard stats daily buckets need aggregate count
- in-memory SQLite tests need StaticPool for background guard logging
- RAG ingest tests need FastAPI dependency overrides for auth
- RagChat has a no-explicit-any lint error

Once #690 lands, I can refresh this branch so its checks rerun against the fixed base.

bajpaisaurabhakoode:
Updated this branch with the shared CI contract fixes that were already passing on the companion AegisAI CI repair branch.

Validation done locally:
- `git diff --check HEAD~1 HEAD`
- `PYTHONPYCACHEPREFIX=/private/tmp/aegis680-pycache python3 -m py_compile backend/app/api/v1/documents.py backend/app/api/v1/guard.py backend/app/core/database.py backend/tests/test_notifications.py backend/tests/test_rag_ingest.py`

I also tried the narrow pytest set for `backend/tests/test_notifications.py backend/tests/test_rag_ingest.py`, but this local temp worktree does not have backend dependencies installed (`sqlalchemy` missing), so GitHub Actions should be the source of truth for the full test run.

PR #681, opened 2026-05-26
fix: move AI Systems filtering server-side

saurabhhhcodes:
## Summary
- add backend search, risk, and compliance filters to GET /api/v1/ai-systems
- update the AI Systems page to send filter params to the API and refetch on filter changes
- reset pagination when filters change and use total pages from the paginated response

## Validation
- PYTHONPYCACHEPREFIX=/private/tmp/aegisai-issue-678/.pycache python3 -m py_compile backend/app/api/v1/ai_systems.py backend/tests/test_ai_systems_filters.py backend/tests/test_frontend_ai_systems_filters.py
- git diff --check
- targeted pytest could not be executed in this local shell because the available interpreters here were missing repo test dependencies such as sqlalchemy or pytest; GitHub Actions should be the final verifier

PR #690, opened 2026-05-26
fix: align CI contracts

saurabhhhcodes:
## Summary
- return 201 from document generation to match the documented create contract and backend tests
- include aggregate count in guard stats daily buckets to satisfy the response schema
- use StaticPool for SQLite in-memory databases so background guard logging sees the same test schema
- authenticate RAG ingest tests with FastAPI dependency overrides instead of only patching the original security symbol
- remove the explicit any from RagChat error handling so frontend lint passes

## Tests
- PYTHONPYCACHEPREFIX=/private/tmp/aegisai-ci-contracts/.pycache python3 -m py_compile backend/app/api/v1/documents.py backend/app/api/v1/guard.py backend/app/core/database.py backend/tests/test_rag_ingest.py
- git diff --check

Note: local pytest is blocked in this temp checkout because the available Python environment is missing sqlalchemy. This PR targets the exact failures reported by GitHub Actions on existing PRs #681/#682/#683.

### DIFF
--- a/backend/app/api/v1/documents.py
+++ b/backend/app/api/v1/documents.py
@@ -241,7 +241,11 @@ def update_document(
     
     return document
 
-@router.post("/generate", response_model=DocumentResponse)
+@router.post(
+    "/generate",
+    response_model=DocumentResponse,
+    status_code=status.HTTP_201_CREATED,
+)
 def generate_document(
     request: DocumentGenerateRequest,
     db: Session = Depends(get_db),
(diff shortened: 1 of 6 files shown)


saurabhhhcodes:
CI is green after the backend test-session follow-up. Current checks: Backend Tests pass, Frontend Lint & Build pass, and repo check pass. This should also unblock the shared CI failures I saw on the related Aegis branches.

PR #702, opened 2026-05-26
feat(rag): add groundedness check to prevent hallucinations (#493)

### LINKED ISSUE #493: [RAG] LLM answers aren't checked against retrieved chunks — hallucination risk
### Motivation

 What I noticed

While going through the RAG module, I realized that after 
`retrieval_chain.py` fetches the chunks and the LLM generates 
an answer — there's nothing that checks whether the answer 
actually came from those chunks.

The LLM just... responds. And we trust it.

## Why I think this is a problem here specifically

Most apps can tolerate occasional hallucinations. AegisAI can't.

People are using this to make actual compliance decisions under 
the EU AI Act. If the RAG module confidently says:

> "You don't need a conformity assessment for this system."

— and that's wrong — the company ships a non-compliant product 
into Europe. That's a real legal consequence.

## What I think we should do

Add a groundedness check after the LLM responds — before we 
return anything to the user.

The idea is simple:
- embed the LLM's answer
- embed the retrieved chunks
- compute cosine similarity between them
- if the answer doesn't match the sources closely enough, 
  flag it as `low_confidence: true`

So the response would look like:

```json
{
  "answer": "...",
  "groundedness_score": 0.76,
  "low_confidence": false,
  "sources": ["chunk 1", "chunk 2"]
}
```

No breaking changes — just extra fields in the response.

RAG Intelligence

## Files I'd touch

- `modules/rag/groundedness.py` — new checker class
- `modules/rag/retrieval_chain.py` — plug the checker in
- `schemas/rag.py` — add the new response fields
- `api/v1/rag.py` — return updated response

Everything needed (numpy, OpenAI embeddings) is already in the stack.

---

I'd like to implement this if you're open to it. 
Happy to discuss before I start.

adityapichikala:
## Summary

Closes #493

This pull request implements a groundedness validation check for the RAG (Retrieval-Augmented Generation) query system to prevent hallucination risks during AI compliance evaluations. It computes a cosine similarity score between the generated LLM response and the retrieved context chunks using `OpenAIEmbeddings`, returning a `groundedness_score` and setting a `low_confidence` flag if the score is below `0.70`.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactor
- [x] Tests
- [ ] Infra / CI

## Checklist

## Screenshots (if UI change)

*No user interface changes were made in this PR.*


adityapichikala:
hi @SdSarthak  ,check it out i am gssoc 2026 contributor

Issue #663, opened 2026-05-25
Upgrade /guard/history pagination from offset-based to cursor-based system


sakeena-7878:
## Motivation:
Offset pagination does not scale well for large datasets and can return inconsistent or slow results when new logs are inserted or when accessing deep pages.

## Proposed solution:
Replace offset pagination with cursor-based pagination using a base64-encoded (scanned_at, id) cursor, enforce stable sorting (scanned_at DESC, id DESC), and return a next_cursor while preserving existing filters (decision, intent, date range).

##Module it touches
LLM Guard

sakeena-7878:
HI @SdSarthak, Please assign this issue to me under Gssoc.

aaniya22:
Hi, I would like to work on this issue under GSSoC'26.
Please assign this to me. Thank you!

PR #704, opened 2026-05-26
Replace offset pagination with cursor-based pagination in guard history

sakeena-7878:
## Summary

Closes... #663
This PR replaces offset-based pagination in the LLM Guard `/history` endpoint with cursor-based pagination using a base64-encoded `(scanned_at, id)` cursor. It improves scalability and ensures stable, consistent results even under heavy data growth or concurrent inserts while preserving existing filters like decision, intent, and date range. Frontend must update accordingly.

---

## Type of Change

* [x] Bug fix
* [x] Refactor

---

## Checklist

* [x] I have read CONTRIBUTING.md
* [x] My code follows the project style 
* [x] I have updated documentation if needed

### DIFF
--- a/backend/app/api/v1/guard.py
+++ b/backend/app/api/v1/guard.py
@@ -15,11 +15,12 @@ from datetime import datetime, timedelta, timezone
 from threading import Lock
 from typing import Optional
 import logging
+import base64
 
 from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException, Query, status
 from fastapi.responses import JSONResponse
 from pydantic import BaseModel
-from sqlalchemy import func
+from sqlalchemy import func, tuple_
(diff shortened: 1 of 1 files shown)


SdSarthak (changes requested):
Three things needed before merge: (1) The existing `PaginatedResponse` schema has required `total` and `page` fields — this PR removes them without updating the shared schema, causing a Pydantic validation error at runtime. Create a separate `CursorPaginatedResponse` schema. (2) Ship the frontend changes together — the current UI still sends a `page` param and expects `total`. (3) Tuple comparison (`ROW(a,b) < (x,y)`) is not supported on SQLite used in tests — add a note or fallback.

sakeena-7878:
Thanks for the review . I will address this.

PR #716, opened 2026-05-26
fix(guard): upgrade /guard/history to cursor-based pagination

### LINKED ISSUE #663: Upgrade /guard/history pagination from offset-based to cursor-based system

aaniya22:
Closes #663

## 📋 Summary

This PR upgrades the `/guard/history` endpoint from offset-based pagination to a cursor-based pagination system, resolving performance and consistency issues with large datasets.

## 🔍 Problem

Offset pagination does not scale well for large datasets and can return inconsistent or slow results when new logs are inserted or when accessing deep pages.

## ✅ Changes Made

- Replaced offset pagination with cursor-based pagination in `backend/app/api/v1/guard.py`
- Used a base64-encoded `(scanned_at, id)` cursor for stable, consistent page traversal
- Enforced stable sorting (`scanned_at DESC, id DESC`)
- Returns a `next_cursor` field while preserving existing filters (decision, intent, date range)

## 📁 Module Affected

- `backend/app/api/v1/guard.py` — LLM Guard history endpoint

### DIFF
--- a/backend/services/slack_notifier.py
+++ b/backend/services/slack_notifier.py
@@ -1,0 +1,99 @@
+"""
+Slack notifier service for SLA breach alerts.
+
+Builds a Slack Block Kit payload for a given helpdesk ticket and
+sends it to a configured Slack webhook URL.
+"""
+
+from __future__ import annotations
+
+import os
+from typing import Any
+
+
+def build_slack_payload(ticket: dict[str, Any]) -> dict[str, Any]:
+    """Build a Slack Block Kit attachment payload for an SLA breach alert.
+
+    Args:
+        ticket: A dict representing the helpdesk ticket. Expected keys
+            (all optional except 'id'):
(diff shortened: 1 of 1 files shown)


SdSarthak:
Closing — the diff contains a `backend/services/slack_notifier.py` file that is a Slack SLA breach alerter for a helpdesk ticketing system, which has no relation to AegisAI. This content appears to be from a different project submitted to the wrong repository by mistake.

PR #718, opened 2026-05-26
fix(ai-systems): move search and filtering to server-side API params

### LINKED ISSUE #678: bug: Client-side filtering only searches current page, not all records
### Describe the bug

In `AISystems.tsx`, the search and filter logic is applied client-side 
on the already-fetched page of results instead of being sent to the 
backend API. This means filters only work within the current 10 items, 
not across all records.

## Current Behaviour

The filtering happens here:

const filteredSystems = systems.filter((system: AISystem) => {
  const matchesSearch = ...
  const matchesRisk = ...
  const matchesCompliance = ...
})

Since `systems` only contains the current page (10 items), searching 
for a system that exists on page 3 will return no results even though 
it exists in the database.

### Steps to reproduce

1. Add more than 10 AI systems
2. Go to page 1
3. Search for a system that exists on page 2 or 3
4. Observe — it shows "No matching AI systems" even though it exists

### Expected behaviour

Search and filter params should be sent to the backend API so all 
records are searched, not just the current page.

### Actual behaviour

Pass `searchTerm`, `riskFilter`, and `complianceFilter` as params 
to `aiSystemsApi.list()`, and also add them to the React Query 
`queryKey` so refetching triggers correctly when filters change.

Files affected:
- `frontend/src/pages/AISystems.tsx`
- `frontend/src/services/api.ts`

### AegisAI version / commit

_No response_

aaniya22:
Closes #678

## 📋 Summary

This PR fixes a bug where search and filter controls in the AI Systems page only searched within the current page of results (10 items) instead of querying all records in the database.

## 🐛 Problem

In `AISystems.tsx`, filtering was applied client-side using `Array.filter()` on the already-fetched page of results. This meant searching for a system on page 2 or 3 while on page 1 would show "No matching AI systems" even though it exists in the database.

## ✅ Fix

**`frontend/src/services/api.ts`**
- Added `search`, `risk_level`, and `compliance_status` as accepted params in `aiSystemsApi.list()`
- These are now forwarded as query params to the backend API

**`frontend/src/pages/AISystems.tsx`**
- Added `searchTerm` to the React Query `queryKey` so refetching triggers correctly when search changes
- Passed `search`, `risk_level`, and `compliance_status` from component state into `aiSystemsApi.list()`
- Removed the client-side `filteredSystems` filter block entirely

## 📁 Files Changed

- `frontend/src/pages/AISystems.tsx`
- `frontend/src/services/api.ts`


SdSarthak (changes requested):
The server-side filtering approach is the right call, but the indentation is broken throughout: the `queryFn` body in `AISystems.tsx` and the `queryParams` object in `api.ts` are both at column 0 inside their enclosing blocks. Please re-indent consistently (2-space indent matching the rest of the file) before merging. The backend counterpart (#681) has correct formatting for reference.

aaniya22:
Fixed the indentation issues in both AISystems.tsx and api.ts as requested. Ready for re-review."

aaniya22:
The CI lint failures are pre-existing issues in Login.tsx, Register.tsx, and api.ts (lines 25 and 334) that are unrelated to this PR's changes. This PR only modifies the aiSystemsApi.list() function and AISystems.tsx query logic. Ready for re-review on the actual changes.

SdSarthak:
This PR looks good — moving AI system filters and search server-side with SQLAlchemy `or_()` is the right approach. However, this branch is currently in conflict with `main`. Please rebase on the latest main and push again.

SdSarthak:
The indentation fix from the earlier review looks good now, but this can't merge as-is: the Frontend Lint & Build check is failing on a real parsing error introduced by the empty-state refactor in `frontend/src/pages/AISystems.tsx`.

ESLint reports: `317:12 error Parsing error: Expected corresponding JSX closing tag for 'h3'`.

Looking at the diff, the old `<EmptyState ... />` component call was replaced with inline JSX, but the replacement left orphaned props from the old component API mixed into the new markup — there's a `message={...}` and `action={...}` block sitting after the inline `<h3>` that don't belong to any JSX element anymore, and the `<h3>` is never properly closed before that dangling content. Can you clean up that block so the JSX tree is well-formed (build a normal `<h3>...</h3>` then a `<p>` for the message, then the conditional button, no leftover `message=`/`action=` props)?

Also worth a look: Backend Tests is failing too (postgres "role 'root' does not exist" during service container startup) — not obviously related to this PR's changes, but flagging in case it's masking something real.

PR #720, opened 2026-05-26
feat(ci): add dedicated pytest workflow for auto-discovery of all tests

### LINKED ISSUE #684: Feature: Add GitHub Actions workflow for test automation
### Motivation

Add a GitHub Actions workflow to automatically discover and execute all pytest test files across different directories pull request.

 
I would like to make a pr under GSSoC'26

### Proposed Solution

```yaml
 name: Run Pytest

on:
  push:
    branches:
      - main
  pull_request:

jobs:
  test:
    runs-on: ubuntu-latest

    steps:
      - name: Checkout code
        uses: actions/checkout@v6

      - name: Setup Python
        uses: actions/setup-python@v5
        with:
          python-version: "3.12"

      - name: Install dependencies
        run: |
          python -m pip install --upgrade pip
          pip install -r requirements.txt

      - name: Run all pytest tests
        run: |
          pytest


Infra / DevOps

### Alternatives Considered

_No response_

aaniya22:
Closes #684

## 📋 Summary

This PR adds a dedicated GitHub Actions workflow for automatically discovering and running all pytest tests across the backend on every push and pull request.

## 🐛 Problem

While the existing `ci.yml` runs backend tests, there was no dedicated workflow focused purely on pytest auto-discovery across all test directories. A standalone workflow makes test results easier to track and debug in the Actions tab.

## ✅ Changes Made

Added `.github/workflows/pytest.yml` which:
- Triggers on push to `main`/`dev` and on all pull requests to `main`
- Spins up a PostgreSQL 15 service for database-dependent tests
- Sets up Python 3.11 and installs backend dependencies
- Runs `pytest --tb=short -q` to auto-discover and execute all test files across the backend directory

## 📁 File Added

- `.github/workflows/pytest.yml`

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,0 +1,50 @@
+name: Run Pytest
+
+on:
+  push:
+    branches: [main, dev]
+  pull_request:
+    branches: [main]
+
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
+jobs:
+  test:
+    name: Pytest (auto-discover all tests)
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:15
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: aegisai_test
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: pip
+
+      - name: Install dependencies
+        working-directory: backend
+        run: pip install -r requirements.txt
+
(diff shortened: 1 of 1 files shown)


SdSarthak (changes requested):
The workflow connects to a Postgres service, but `conftest.py` unconditionally overrides `DATABASE_URL` to `sqlite:///:memory:` before any imports. The Postgres service will spin up but remain unused and the tests will still run on SQLite. Two ways to fix: (1) update `conftest.py` to respect an already-set `DATABASE_URL` (use `os.environ.setdefault` instead of direct assignment), or (2) remove the Postgres service and use SQLite in the workflow. Pick whichever direction you intend and make them consistent.

SdSarthak (changes requested):
The pytest auto-discovery workflow is a good addition but has a conflict with the existing test configuration: the workflow spins up a Postgres service container, but `backend/tests/conftest.py` uses SQLite for the in-memory test database. Running pytest in the new workflow will fail because the Postgres service is unused and the existing conftest fixture creates an in-memory SQLite session.

Please either (a) update `conftest.py` to use the Postgres service defined in the workflow, or (b) remove the Postgres service block from the workflow and rely solely on the existing SQLite setup.

PR #741, opened 2026-05-27
fix(frontend): handle paginated notifications response

disha-bose-8:
## Summary

Fixes frontend notification handling for paginated API responses.

The notifications endpoint returns:
{
  "items": [],
  "total": 0,
  "page": 1,
  "limit": 50
}

but NotificationBell expected a plain array, which could break unread notification rendering.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## Changes Made

- Updated NotificationBell to read notifications from `data?.items ?? []`
- Preserved unread filtering logic
- Verified backend notifications API response structure in Swagger

## Testing

- [x] Frontend loads correctly
- [x] Notifications endpoint tested via Swagger auth
- [x] No notification rendering crash

### DIFF
--- a/frontend/src/components/NotificationBell.tsx
+++ b/frontend/src/components/NotificationBell.tsx
@@ -48,14 +48,19 @@ export default function NotificationBell() {
   const wrapperRef = useRef<HTMLDivElement>(null)
 
   // Live data via useQuery
-  const queryClient = useQueryClient()
-  const { data: notifications = [] } = useQuery({
-    queryKey: ['notifications', 'unread'],
-    queryFn: () => notificationsApi.list(true),
-    refetchInterval: 60_000,
-  })
+const queryClient = useQueryClient()
 
-  const unreadCount = notifications.filter((n: NotificationPreview) => !n.is_read).length
+const { data } = useQuery({
+  queryKey: ['notifications', 'unread'],
+  queryFn: () => notificationsApi.list(true),
+  refetchInterval: 60_000,
+})
+
+const notifications = data?.items ?? []
+
+const unreadCount = notifications.filter(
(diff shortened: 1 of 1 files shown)


SdSarthak (changes requested):
The logic is correct (handling `data?.items` vs a plain array) but the indentation inside `NotificationBell.tsx` is broken — the `useQuery` call block is at column 0 while the surrounding code is indented. Please re-indent consistently with the rest of the component before merging.

SdSarthak (changes requested):
The notification response handling fix is correct and needed, but the indentation in `NotificationBell.tsx` is still broken — JSX elements inside the conditional are at the wrong nesting level, which will cause rendering issues. Please fix the indentation throughout the component and re-submit.

SdSarthak (changes requested):
The logic fix here is correct — extracting `data?.items ?? []` instead of treating the full paginated response as an array is exactly right.

However, the new code has broken indentation: the `const queryClient`, `useQuery`, `const notifications`, and `const unreadCount` declarations are at column 0 instead of the 2-space indent used by the surrounding function body. This will fail the frontend lint/build step.

Please rebase and re-submit with consistent 2-space indentation matching the rest of the component.

SdSarthak:
This branch is currently in conflict with main. Please rebase on the latest main and push again — the paginated notifications handling fix (data?.items ?? []) is the right approach.

SdSarthak:
To be clear — beyond the rebase, this still needs the indentation fix from the earlier review (useQuery/notifications/unreadCount declarations should match the surrounding 2-space indent in NotificationBell.tsx).

PR #742, opened 2026-05-27
Fix/cursor pagination

### LINKED ISSUE #663: Upgrade /guard/history pagination from offset-based to cursor-based system

sakeena-7878:
## Summary

Closes #663
This PR replaces offset-based pagination in the LLM Guard /history endpoint with cursor-based pagination using a base64-encoded (scanned_at, id) cursor.

A dedicated CursorPaginatedResponse schema is introduced to avoid breaking the existing PaginatedResponse contract used elsewhere in the system.
SQLite compatibility handled (no tuple comparison operator used; logic uses explicit and_/or_ filtering)
Frontend has been updated to support cursor-based pagination instead of page/offset.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Refactor

## Checklist

## Screenshots (if UI change)
Replaced page/limit pagination with cursor/next_cursor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,15 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ---
 
+## [Unreleased]
+
+- **Fixed** Guard API merge conflicts and resolved pagination inconsistencies in history endpoint
+- **Changed** Updated frontend Guard/RAG API types (removed duplicate interfaces, improved type safety)
+- **Fixed** ESLint issues in frontend services and components
+- **Changed** Improved cursor-based pagination handling for guard scan history
+
+---
+
 ## [0.1.0] — 2026-04-05
 
 ### Added
--- a/backend/app/api/v1/guard.py
+++ b/backend/app/api/v1/guard.py
@@ -10,21 +10,16 @@ TODO for contributors (medium difficulty):
 """
 
 import hashlib
+import logging
+import base64
+
 from collections import Counter, defaultdict, deque
 from datetime import datetime, timedelta, timezone
 from threading import Lock
 from typing import Optional, TypedDict
 
 from app.api.v1.webhooks import deliver_webhook
-import logging
-import base64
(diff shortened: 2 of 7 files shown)


SdSarthak (changes requested):
Two regressions need to be fixed before this can merge:

1. The empty-prompts validation was removed (`if not self.prompts: raise ValueError`). Batch requests with zero prompts will now silently succeed with an empty response — please restore that guard.
2. `db: Session` was removed from `scan_prompt` along with the webhook delivery block. That coupling was intentional — the webhook trigger relies on the db session. If the goal is to separate concerns, use `BackgroundTasks` to deliver the webhook asynchronously, but don't drop it entirely.

The `CursorPagination` helper itself and the `CursorPaginatedResponse` schema are the right direction — just fix the two regressions above.

sakeena-7878:
Ok got it

SdSarthak:
The two regression fixes are in — the empty-prompts guard is restored and the `db: Session` + webhook delivery block are back. The implementation now looks correct. Please rebase on `main` to resolve the current conflicts and it will be ready to merge.

sakeena-7878:
This PR resolves guard module merge conflicts and completes the cursor-based pagination refactor.

Key updates:
- Backend: cleaned guard endpoints, fixed schema imports, resolved pagination logic conflicts
- Frontend: fixed API typing issues (RagSource, Guard interfaces), removed duplicates
- Lint/build: frontend passes lint and production build successfully
- Tests: updated and passing for guard API and related modules

Please let me know if any adjustments are needed.

sakeena-7878:
Thanks.

PR #750, opened 2026-05-27
fix: include answer_id in answer state to show RAG feedback buttons

### LINKED ISSUE #738: RAG chat never shows feedback buttons because answer_id is dropped
Noticed this while testing the RAG chat UI: the thumbs up/down feedback controls never show up after an answer comes back.

The backend response includes an `answer_id`, and the UI already checks for it before rendering the feedback block. But in `frontend/src/pages/RagChat.tsx`, the response is copied into state like this around line 74:

```ts
setAnswer({
  answer: data.answer,
  sources: data.sources || [],
})
```

So `data.answer_id` gets dropped on the floor. Later the component does this:

```tsx
{answer.answer_id && (...feedback buttons...)}
```

Since `answer_id` is never stored, that condition is always false and the buttons stay hidden forever.

This looks like a small frontend wiring fix: include `answer_id: data.answer_id` when setting the answer state, and make sure the answer type includes that field.

eshaanag:
## What does this PR do?

Fixes #738

## Problem
The RAG chat response from the backend includes an answer_id field,
but it was being dropped when setting the answer state in RagChat.tsx.
The feedback buttons check for answer.answer_id before rendering,
so they were permanently hidden since the value was never stored.

## Fix
Added answer_id: data.answer_id to the setAnswer call and updated
the answer type to include the optional answer_id field.

## Files changed
- frontend/src/pages/RagChat.tsx

## How to test
1. Open RAG chat and ask any question
2. After the answer loads, feedback buttons should now appear
3. Clicking thumbs up/down should work correctly

### DIFF
--- a/frontend/src/pages/RagChat.tsx
+++ b/frontend/src/pages/RagChat.tsx
@@ -74,6 +74,7 @@ export default function RagChat() {
       setAnswer({
         answer: data.answer,
         sources: data.sources || [],
+        answer_id: data.answer_id,
       })
     } catch (err: unknown) {
       // ✅ ERROR HANDLING


eshaanag:
Hi @maintainer, this PR was raised as part of GSSoC 2026 contribution.  
Could you please add the appropriate GSSoC label if applicable? Thanks!

eshaanag:
Quick validation update: I checked this branch locally and `npm run build` passes successfully.

Build output completed with Vite production build; the only note is the existing large chunk warning, which is unrelated to this one-line fix.

PR #774, opened 2026-05-28
fix: add max-length validation for full_name and company_name fields

sampawar27:
# Summary

Closes #<issue-number>

This PR adds max-length validation for `full_name` and `company_name` fields to improve input validation and prevent oversized values from being stored in the database.

### Changes Made

* Updated `full_name` and `company_name` field lengths from `255` to `100`
* Added validation tests in `backend/tests/test_auth.py`
* Updated `tsconfig.json` `ignoreDeprecations` value to `6.0`

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Documentation update
* [ ] Refactor
* [x] Tests
* [ ] Infra / CI

## Checklist

* [x] I have read CONTRIBUTING.md
* [x] My code follows the project style
* [x] I have added/updated tests where relevant
* [ ] pytest backend/tests/ passes locally
* [x] I have not committed .env or any secrets

## Screenshots (if UI change)

N/A

### DIFF
--- a/backend/app/models/user.py
+++ b/backend/app/models/user.py
@@ -18,8 +18,8 @@ class User(Base):
     id = Column(Integer, primary_key=True, index=True)
     email = Column(String(255), unique=True, index=True, nullable=False)
     hashed_password = Column(String(255), nullable=False)
-    full_name = Column(String(255))
-    company_name = Column(String(255))
+    full_name = Column(String(100))
+    company_name = Column(String(100))
 
     # Subscription
     subscription_tier = Column(Enum(SubscriptionTier), default=SubscriptionTier.FREE)


SdSarthak (changes requested):
Great direction — this aligns the model-level validation with the Pydantic schema changes from #747. However, shortening an existing VARCHAR column (255 → 100) is a **schema migration** that requires an Alembic migration file.

Please add a migration under `backend/alembic/versions/` that runs `op.alter_column('users', 'full_name', existing_type=sa.String(255), type_=sa.String(100))` (and same for `company_name`). Without the migration, existing databases on the `255` schema will not pick up the constraint change and could store data that violates the new model definition.

Once the migration file is included this is good to go.

PR #776, opened 2026-05-28
Feature/copy button reports

Priyam-1407:
## What does this PR do?

* Added copy button for generated compliance reports
* Added toast notification after copying content
* Added temporary check icon feedback after successful copy
* Resolved merge conflicts in `Documents.tsx`


SdSarthak (changes requested):
Two issues that need fixing before this can merge:

1. **Duplicate `handleCopy` function**: The existing `handleCopy(docId: number, content: string)` at the top of the component is left in place while a second `handleCopy(doc: Document)` is added at line ~128. TypeScript will throw a 'Duplicate identifier' compile error. Remove the old definition and keep only the new one with the `doc: Document` signature.

2. **Broken button JSX**: The copy button's `<button` opening tag and its attributes are at inconsistent indentation levels, and the `CopyButton` import is still present but all uses have been removed, causing an 'unused import' warning.

Fix both and this is mergeable.

SdSarthak (commented):

Priyam-1407:
Hey @SdSarthak, fixed both issues:

- Removed the duplicate `handleCopy` function — the old one was referencing `setCopiedDocId` which didn't exist. Kept the clean version that uses `setCopiedId` + toast notification.
- Added missing `currentPage` and `limit` state variables for pagination.
- Branch is now rebased onto main, conflicts resolved.

Please re-review when you get a chance!

SdSarthak:
This still conflicts with main. Could you rebase onto origin/main, resolve the conflicts, and force-push when you get a chance?

PR #803, opened 2026-05-29
Add Guard audit logging and /api/v1/guard/audit-log endpoint

Pratha2005:
## Summary

Closes #

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor
- [ ] Tests
- [ ] Infra / CI

## Checklist

- [ ] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [ ] `pytest backend/tests/` passes locally


SdSarthak (changes requested):
The audit log endpoint idea is good, but there's a structural issue: a separate `guard_audit_logs` table duplicates data already stored in `guard_scan_logs`. The existing table has `prompt_hash`, `decision`, `confidence`, `detection_type`, and `scanned_at` columns — all the fields needed for an audit view.

The right approach is to expose an endpoint that reads directly from `GuardScanLog`, filtered by `user_id`. This avoids a second write on every scan, removes the data duplication, and eliminates a migration for a new table.

Additionally, `GuardAuditLog` is imported twice in `models/__init__.py`.

Please revise to:
1. Remove `GuardAuditLog` model, schema, and migration
2. Remove the `_build_guard_audit_log` helper and second `db.add()` call in `log_scan`  
3. Implement `GET /guard/audit-log` reading from the existing `GuardScanLog` model

The endpoint signature (pagination, decision filter, date range) looks correct — just wire it to `GuardScanLog` instead.

PR #814, opened 2026-05-29
fix: handle empty API response with fallback error state (fixes #664)

### LINKED ISSUE #664: Bug: Empty API response causes UI freeze instead of fallback error state
### Describe the bug

When the backend/API returns an empty response (null, empty object, or unexpected blank payload), the frontend UI becomes unresponsive instead of displaying a fallback error state or retry option.

### Steps to reproduce

Open the application.
Trigger an API request that returns an empty or invalid response.
Observe the frontend behavior after response handling.

### Expected behaviour

The UI should gracefully handle empty API responses by:

Showing an error message / fallback state
Preventing UI freeze
Allowing retry or further interaction

### Actual behaviour

The application UI freezes / loading state persists, and the user cannot proceed normally.

**Possible Cause**
Improper null/undefined response handling in API response parsing or state update logic.

**Suggested Fix**

Add validation checks before rendering/updating UI state:

Handle null / undefined / empty payloads
Display fallback error component
Reset loading state properly

### AegisAI version / commit

Latest main branch (current repository state)

BCA-krishna:
## Problem
When the API returned an empty response, the UI would freeze 
instead of showing a fallback error state.

## Changes Made
- Added `assertNonEmpty()` helper to validate API responses
- Applied empty response checks to `classificationApi`, `ragApi`, and `guardApi`
- Throws a clear error message instead of letting UI freeze

## Testing
- Empty response now throws an error that can be caught by the UI
- Normal responses work as before

Closes #664

### DIFF
--- a/frontend/src/services/api.ts
+++ b/frontend/src/services/api.ts
@@ -1,6 +1,17 @@
 import axios from 'axios'
 import { useAuthStore } from '../stores/authStore'
 
+// Fix for Issue #664: Validate API response is not empty
+function assertNonEmpty<T>(data: T, endpoint: string): T {
+  if (data === null || data === undefined) {
+    throw new Error(`Empty response received from ${endpoint}`)
+  }
+  if (typeof data === 'object' && Object.keys(data).length === 0) {
+    throw new Error(`Empty response received from ${endpoint}`)
+  }
+  return data
+}
+
 const api = axios.create({
   baseURL: '/api/v1',
   headers: {
@@ -104,11 +115,13 @@ export const aiSystemsApi = {
 export const classificationApi = {
   classify: async (data: Record<string, unknown>) => {
     const response = await api.post('/classification/classify', data)
-    return response.data
+        return assertNonEmpty(response.data, '/classification/classify')
+
   },
   classifyAndSave: async (systemId: number, data: Record<string, unknown>) => {
     const response = await api.post(`/classification/classify/${systemId}`, data)
-    return response.data
+        return assertNonEmpty(response.data, `/classification/classify/${systemId}`)
+
   },
 }
 
@@ -164,7 +177,8 @@ export const ragApi = {
     const { data } = await api.post('/rag/query', {
       question,
     })
-    return data
+        return assertNonEmpty(data, '/rag/query')
+
   },
   feedback: async (payload: { answer_id: string; vote: 'up' | 'down' }) => {
     const { data } = await api.post('/rag/feedback', {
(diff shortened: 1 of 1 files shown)


SdSarthak (changes requested):
The `assertNonEmpty` helper is the right approach, but the implementation has indentation bugs that will cause TypeScript errors:

In `classificationApi.classify`, `classificationApi.classifyAndSave`, and `ragApi.query`, the `return assertNonEmpty(...);` lines are indented with 8 spaces instead of the surrounding 4-space indent. TypeScript will flag these as syntax errors. Please fix the indentation to match the surrounding code.

Also: removing the `normalizeSources` function from `RagChat.tsx` is a regression — it handles the case where sources from the API are strings instead of objects. Without it, `sources: data.sources || []` will pass raw strings to the component expecting `{ title, excerpt }` objects. Please keep the `normalizeSources` function or ensure the API always returns structured objects before removing it.

SdSarthak (commented):

SdSarthak (changes requested):
The intent here — returning an error when the API gives back an empty response — is good. However there are two issues:

1. **Broken indentation**: the `return assertNonEmpty(...)` calls use 8-space indentation (double the 4-space surrounding context), making the code inconsistent and causing potential lint failures. Please align to the surrounding 4-space indent in `api.ts`.
2. **Checking for empty objects is too broad**: `Object.keys(data).length === 0` would throw on a valid empty list endpoint (e.g., a paginated response with zero items that returns `{items: [], total: 0}`). Consider narrowing the check to `data === null || data === undefined` only, or scoping it to specific fields you actually need.

Please fix the indentation and narrow the validation logic, then rebase on the latest main.

SdSarthak:
This branch is currently in conflict with main. Please rebase on the latest main and push again.

SdSarthak:
To be clear — beyond the rebase, this still needs the two fixes from the earlier reviews: fix the 8-space indentation on the assertNonEmpty return lines in api.ts, and either restore normalizeSources in RagChat.tsx or otherwise handle string-typed sources so the component doesn't break on non-object entries.

SdSarthak:
Checking the current state of this branch: the diff against main is now empty (0 files changed) after the 2026-07-08 merge commit, so the empty-response handling and normalizeSources fix from this PR aren't present anymore, on top of the indentation and normalizeSources concerns from the earlier reviews never having been addressed. Please rebase against latest main so the actual changes show up again, and fix the two outstanding issues: the assertNonEmpty indentation in api.ts and the normalizeSources handling in RagChat.tsx.

PR #817, opened 2026-05-29
fix: avoid duplicate unauthenticated document save

### LINKED ISSUE #751: Document editor save path drops auth headers from Documents page
I ran into this while checking why document edits were not persisting from the Documents page.

There is a save handler in `frontend/src/pages/Documents.tsx` that calls the protected document update endpoint with raw `fetch()`:

```ts
const response = await fetch(`/api/v1/documents/${editingDoc.id}`, {
  method: 'PUT',
  headers: {
    'Content-Type': 'application/json',
  },
  body: JSON.stringify({ content })
})
```

This is around `frontend/src/pages/Documents.tsx:111`.

Because this bypasses the shared API client, the auth token is never attached. The backend document routes are protected, so the save request can come back as `401 Unauthorized`. From the user side, it looks like the editor just quietly failed because the catch block only logs to the console.

This is separate from the editor component save path. The fix should be to route this through the existing authenticated API service, or at least attach the same Authorization header/interceptor behavior used everywhere else. It would also be good to surface the failure in the UI instead of only `console.error`.

eshaanag:
## Summary

Closes #751

This PR removes the duplicate raw fetch save from the Documents page after the editor saves a document. The DocumentEditor already writes through the shared authenticated API client, so the page-level callback now only refreshes the documents query after a successful editor save.

That keeps the save path on the configured API client, avoids a second unauthenticated PUT request, and lets the editor continue showing its existing save error state when the authenticated save fails.

Hi @maintainer, this PR was raised as part of a GSSoC 2026 contribution for #751. Could you please add the appropriate GSSoC label if applicable? Thanks!

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor
- [ ] Tests
- [ ] Infra / CI

## Checklist

- [x] My code follows the project style
- [x] npm run lint passes locally
- [x] npm run build passes locally
- [x] I have not committed .env or any secrets

## Local validation

- npm run lint
- npm run build
- git diff --check


eshaanag:
Hi @maintainer, this PR was raised as part of a GSSoC 2026 contribution for #751. Could you please add the appropriate GSSoC label if applicable? Thanks!

eshaanag:
Validation update: all GitHub checks are passing now — Backend Tests, Frontend Lint & Build, and Changelog Check.

This should be ready for review under #751.

SdSarthak (changes requested):
The new `handleSaveDocument` is a regression — it removes the API call and only invalidates the query cache, so changes made in the editor are never actually saved. The document content passed as the argument is silently discarded.

The original issue (unauthenticated save) should be fixed by using the authenticated `api` Axios instance (which already injects the Authorization header) instead of a bare `fetch`. Please replace the `fetch` call with:

```ts
const handleSaveDocument = async (content: string) => {
  if (!editingDoc) return
  try {
    await api.put(`/documents/${editingDoc.id}`, { content })
    queryClient.invalidateQueries({ queryKey: ['documents'] })
  } catch {
    toast.error('Save failed')
  }
}
```

`api` is the Axios instance from `../services/api` — it already attaches the JWT token via the request interceptor.

eshaanag:
Pushed the requested update. The save path now uses the authenticated `api.put('/documents/:id')` call and shows a visible toast error on failure.\n\nOne extra note: after rebasing on current `main`, the frontend build was failing in `RagChat.tsx` because `ragApi.query` types `sources` as `Array<string | { title, excerpt }>` while the page state expects structured source objects. I added a small normalization helper there so `npm run build` passes on the rebased branch.\n\nValidated locally with `npm run lint` and `npm run build` from `frontend/`.

SdSarthak:
Superseded by #881 which implements the authenticated document save correctly.

PR #826, opened 2026-05-29
Fix/login auth race condition

### LINKED ISSUE #306: fix(frontend): getMe() called before token is stored in auth state — potential 401 on fresh login
### Describe the bug

In Login.tsx, the token returned from authApi.login() is not stored 
in the Zustand auth store before authApi.getMe() is called. Since the 
axios interceptor reads the Bearer token from the store, getMe() fires 
without authentication and returns 401 Unauthorized on fresh sessions.

This causes login to silently fail — the user appears to be logged out 
even though credentials are correct.

### Steps to reproduce

1. Run `docker compose up -d`
2. Register a new account
3. Try to log in
4. Login returns 200 but getMe returns 401 — user stays logged out

### Expected behaviour

After login, the token should be stored in the auth store before any 
authenticated API calls are made, ensuring getMe() always has a valid Bearer token.

### Actual behaviour

In Login.tsx, getMe() is called before setAuth() stores the token.
The axios interceptor reads the token from Zustand store, but the store 
is empty at the time getMe() fires — this causes a 401 on fresh sessions 
where no cached token exists.

Backend logs confirm:
POST /api/v1/auth/login  200 OK
GET  /api/v1/auth/me     401 Unauthorized

### AegisAI version / commit

_No response_

shivanisri10:
## Summary

Closes #306

This PR fixes a login race condition where `getMe()` was called before the authentication token was stored in the Zustand auth store. This caused the axios interceptor to send unauthenticated requests and resulted in 401 errors on fresh logins. The fix ensures the token is stored before any authenticated API calls are made, allowing `getMe()` to work correctly after login.

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Documentation update
* [ ] Refactor
* [ ] Tests
* [ ] Infra / CI

## Checklist

* [x] I have read [[CONTRIBUTING.md](https://chatgpt.com/CONTRIBUTING.md)](../CONTRIBUTING.md)
* [x] My code follows the project style (ESLint for TS)
* [ ] I have added/updated tests where relevant
* [x] `pytest backend/tests/` passes locally

## Screenshots (if UI change)

Not applicable

### DIFF
--- a/frontend/src/pages/Login.tsx
+++ b/frontend/src/pages/Login.tsx
@@ -21,20 +21,16 @@ export default function Login() {
     try {
       const tokenData = await authApi.login(email, password)
 
-      // 1. Immediately store token
-      setAuth(tokenData.access_token, null)
+setAuth(tokenData.access_token, null)
 
-      // 2. FORCE attach token for immediate next request
-      axios.defaults.headers.common[
-        'Authorization'
-      ] = `Bearer ${tokenData.access_token}`
+axios.defaults.headers.common['Authorization'] =
+  `Bearer ${tokenData.access_token}`
 
-      // 3. Now safely call getMe
-      const user = await authApi.getMe()
+const user = await authApi.getMe()
 
-      setAuth(tokenData.access_token, user)
+setAuth(tokenData.access_token, user)
 
-      navigate('/')
+navigate('/')
     } catch (err) {
       if (axios.isAxiosError(err) && err.response?.data?.detail) {
         setError(err.response.data.detail)
--- a/frontend/src/pages/Login.tsx
+++ b/frontend/src/pages/Login.tsx
@@ -20,9 +20,20 @@ export default function Login() {
 
     try {
       const tokenData = await authApi.login(email, password)
-      setAuth(tokenData.access_token, null)  
-      const user = await authApi.getMe()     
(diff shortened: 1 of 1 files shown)


SdSarthak (changes requested):
Setting `axios.defaults.headers.common['Authorization']` globally modifies a shared object — this affects all Axios instances and can leak the token to third-party requests. The `api` instance already attaches the token via its request interceptor; there's no need for this extra line.

Also the indentation inside the `try` block is broken — lines 21-28 are at column 0 instead of being indented inside the `try` block. This will cause a syntax error.

Please remove the `axios.defaults.headers.common` line (the interceptor handles this correctly) and fix the indentation. Once those two issues are fixed this is ready to merge.

SdSarthak (commented):

SdSarthak:
Fixing the login auth race condition is a solid bug fix. This branch is currently in conflict with `main`. Please rebase on the latest main and push again.

SdSarthak (changes requested):
The lines added inside the `try` block (`setAuth(tokenData.access_token, null)`, `axios.defaults.headers...`, etc.) are at 0-indent — they need to be indented consistently with the surrounding code. Please fix the indentation.

PR #837, opened 2026-05-30
Improve scan prompt tests with payload verificationImprove scan prompt tests with payload verification

cosmoqain459:
## Summary

This PR improves the `scan_prompts` test suite by verifying that prompt content is correctly sent to the guard scanning API.

### Changes

* Added assertions to verify HTTP requests are made exactly once.
* Added payload verification checks for both malicious and safe prompt scenarios.
* Preserved existing validation for exit codes and report generation.

### Benefits

* Improves test reliability.
* Validates request composition rather than only mocked responses.
* Strengthens coverage for the prompt scanning workflow.


SdSarthak (changes requested):
Small improvements to the test assertions, but two things to fix:

**1. Missing trailing newline (twice)**

The file now ends with `\ No newline at end of file` — both the malicious and clean test functions end without a final newline. Please add a trailing newline to the end of the file.

**2. Unused variable**

`args, kwargs = client.post.call_args` — `args` is assigned but never used. Either use `_, kwargs = client.post.call_args` or just `kwargs = client.post.call_args.kwargs`.

The approach of asserting on `assert_called_once()` and checking the payload content is good.

PR #847, opened 2026-05-30
feat: add PATCH /api/v1/ai-systems/{id}/status endpoint

### LINKED ISSUE #24: Add PATCH /api/v1/ai-systems/{id}/status endpoint
Add a targeted endpoint to update only the `compliance_status` of an AI system — without requiring a full PUT of all fields. Useful for marking a system compliant after documentation is complete.

**File:** `backend/app/api/v1/ai_systems.py`

Bhavy12-cell:
Closes #24

Added a targeted PATCH endpoint to update only the compliance_status 
of an AI system without requiring a full PUT of all fields.

Changes made:
- Added PATCH /{system_id}/status endpoint in ai_systems.py
- Uses ComplianceStatusUpdateSchema for request validation
- Validates ownership before updating
- Returns 404 if system not found or not owned by current user
- Returns updated AISystemResponse after successful update


SdSarthak (changes requested):
The PR description says this adds a PATCH `/{system_id}/status` endpoint to `ai_systems.py`, but the diff doesn't include any changes to `backend/app/api/v1/ai_systems.py`. The files that are changed (`main.py`, `requirements.txt`, `Documents.tsx`, `api.ts`) don't implement what's described.

Specifically:
- `main.py` adds `scheduler.shutdown()` in the lifespan handler, but there is no `scheduler` imported or initialized anywhere in scope — this would throw a `NameError`
- `requirements.txt` adds `apscheduler` which is also unrelated to a status PATCH endpoint
- The Documents.tsx and api.ts changes add pagination to the documents list, which is a separate concern

It looks like the wrong branch may have been pushed. Please check which branch this PR targets and make sure the actual PATCH endpoint changes are included.

PR #863, opened 2026-05-31
fix: validate RAG feedback vote values

### LINKED ISSUE #802: RAG feedback accepts invalid vote values and treats them as thumbs-down
### Describe the bug

Hi @SdSarthak,

I found a validation issue in the RAG feedback endpoint.

The endpoint is intended to accept only two feedback values: up and down. However, the current request handling accepts the vote as a plain string. The endpoint only checks whether the vote is up; every other value is treated as a thumbs-down vote.

This means values such as maybe, neutral, invalid, or even a typo can be accepted and counted as negative feedback. Over time, this could pollute feedback data and affect the low-quality chunk reporting logic.

Possible approach
The request model can be updated so the vote field only allows up or down. This would let request validation reject invalid values before any feedback counts are changed.

Suggested test coverage:
Add tests to confirm that an up vote increments thumbs-up, a down vote increments thumbs-down, and an invalid vote such as maybe is rejected with a validation error.

### Steps to reproduce

Start the backend locally.
Log in and obtain a valid authentication token.
Submit a RAG query through the RAG chat page or the /api/v1/rag/query endpoint.
Copy the returned answer_id.
Send feedback to the RAG feedback endpoint using that answer_id.
Use an invalid vote value such as maybe, neutral, or invalid.
Observe that the request is accepted instead of rejected.
The invalid value is treated as a thumbs-down vote.

### Expected behaviour

The RAG feedback endpoint should only accept up and down as valid vote values. Any other value should return a validation error instead of changing feedback counts.

### Actual behaviour

The endpoint accepts any string as the vote value. If the value is exactly up, it increments the thumbs-up count. For every other value, including invalid values like maybe, neutral, or misspellings, it increments the thumbs-down count.

### AegisAI version / commit

_No response_

abhishekbidare03:
## Summary

The RAG feedback endpoint accepted arbitrary vote values and treated every value other than "up" as a thumbs-down vote.

This change restricts accepted vote values to "up" and "down" using request validation.

## Changes

- Added strict validation for feedback vote values
- Invalid values now return HTTP 422
- Added regression test covering invalid vote submissions

## Verification

- Valid votes ("up", "down") are accepted
- Invalid votes return HTTP 422
- Invalid votes do not affect low-quality chunk metrics

Closes #802

### DIFF
--- a/backend/app/api/v1/rag.py
+++ b/backend/app/api/v1/rag.py
@@ -13,7 +13,7 @@ import os
 import shutil
 import tempfile
 import time
-from typing import List
+from typing import List, Literal
 
 from fastapi import APIRouter, Depends, File, HTTPException, UploadFile, status
 from pydantic import BaseModel
@@ -233,7 +233,7 @@ def rag_health():
 
 class RAGFeedbackRequest(BaseModel):
     answer_id: str
-    vote: str  # "up" or "down"
+    vote: Literal["up", "down"]
 
 
 @router.post("/feedback")
--- a/backend/tests/integration/test_rag_feedback.py
+++ b/backend/tests/integration/test_rag_feedback.py
@@ -15,6 +15,20 @@ class DummyDoc:
         self.metadata = {"source": source}
 
 
+def _install_fake_rag_chain(fake_result):
+    """Install a lightweight fake retrieval chain for RAG endpoint tests."""
+    import sys
+    import types
+
+    mod = types.ModuleType("app.modules.rag.retrieval_chain")
+
+    def _fake_get_qa_chain():
+        return lambda payload: fake_result
+
+    mod.get_qa_chain = _fake_get_qa_chain
+    sys.modules["app.modules.rag.retrieval_chain"] = mod
+
+
 def _get_test_db():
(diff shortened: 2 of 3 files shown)


SdSarthak (commented):
Good fix — using `Literal["up", "down"]` is the right approach for constraining the vote field at the Pydantic layer. CI is failing because rag.py has been updated by other merged PRs. Please rebase on main and push.

abhishekbidare03:
Thanks for the review.

I've rebased the branch on the latest main and pushed the updated changes.

The remaining CI failure appears to be coming from `frontend/src/pages/RagChat.tsx` due to a TypeScript type mismatch (`(string | RagSource)[]` vs `RagSource[]`). My changes for this PR are limited to the RAG feedback vote validation and associated backend test.

Please let me know if you would like me to investigate the frontend type issue as part of this PR, or if it should be handled separately.

abhishekbidare03:
Thanks for the review @SdSarthak .

I've rebased on the latest main and pushed the requested updates. The frontend type issue has also been addressed.

Please let me know if there are any further changes required.

PR #948, opened 2026-06-03
fix: handle duplicate name collision in AI system clone endpoint

### LINKED ISSUE #795: [BUG] Unhandled Database IntegrityError in AI System Cloning due to Name Collisions
### Describe the bug

A database crash occurs when a user tries to clone the same AI system multiple times, or clones an AI system that already has a copy. This is because the database enforces a unique constraint on `(owner_id, name)`, but the cloning endpoint unconditionally appends ` (copy)` to the original name and commits it without check or exception handling.

### Steps to reproduce

1. Authenticate and create a new AI system named "Alpha".
2. Trigger the clone endpoint `POST /api/v1/ai-systems/{system_id}/clone`. A new system named "Alpha (copy)" is created successfully.
3. Trigger the clone endpoint for "Alpha" again.
4. Observe the response from the server.

### Expected behaviour

The clone endpoint should dynamically check if the name "Alpha (copy)" already exists for the user. If it does, it should resolve the collision by incrementing a suffix (e.g., "Alpha (copy 2)", "Alpha (copy 3)") to avoid database-level crashes and register the clone successfully.

### Actual behaviour

The server crashes with an unhandled database exception and returns a `500 Internal Server Error` to the client. The backend log displays:

### AegisAI version / commit

Main Branch (Latest Commit) Affected File: `backend/app/api/v1/ai_systems.py` (Line 474-505)

Gungun210:
Fixes #795

## Problem
When cloning an AI system multiple times, the endpoint would crash with 
an IntegrityError because "Alpha (copy)" already existed.

## Solution
- Added a while loop to check if the clone name already exists
- If "Alpha (copy)" exists, tries "Alpha (copy 2)", "Alpha (copy 3)" etc.
- Added try/except IntegrityError as a safety net

## Testing
- Clone same system twice → should create "Alpha (copy)" and "Alpha (copy 2)"

### DIFF
--- a/backend/app/api/v1/ai_systems.py
+++ b/backend/app/api/v1/ai_systems.py
@@ -446,18 +446,36 @@ def clone_ai_system(
             detail="AI system not found"
         )
 
-    cloned = AISystem(
-        owner_id=current_user.id,
-        name=f"{original.name} (copy)",
-        description=original.description,
-        version=original.version,
-        use_case=original.use_case,
-        sector=original.sector,
-        compliance_status=ComplianceStatus.NOT_STARTED,
-    )
+        base_name = f"{original.name} (copy)"
+        clone_name = base_name
+        counter = 2
 
-    db.add(cloned)
-    db.commit()
+        while db.query(AISystem).filter(
+            AISystem.owner_id == current_user.id,
+            AISystem.name == clone_name,
+        ).first():
+            clone_name = f"{original.name} (copy {counter})"
+            counter += 1
+
+        cloned = AISystem(
+            owner_id=current_user.id,
+            name=clone_name,
+            description=original.description,
+            version=original.version,
+            use_case=original.use_case,
+            sector=original.sector,
+            compliance_status=ComplianceStatus.NOT_STARTED,
+        )
+
+        db.add(cloned)
+    try:
+        db.commit()
+    except IntegrityError:
+        db.rollback()
+        raise HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail="Could not create a unique clone name.",
+        )
     db.refresh(cloned)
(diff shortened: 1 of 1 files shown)


SdSarthak (changes requested):
The incremental suffix logic (`(copy)`, `(copy 2)`, `(copy 3)`, …) is the right approach. However, the indentation in the diff is off — the `base_name`, `clone_name`, `counter` assignments and the `while` loop appear indented into a block that doesn't exist, and the `cloned` assignment ends up outside the try block while the `db.add` lands in an inconsistent position.

Please clean up the indentation so the entire name-resolution loop and object creation happens before the `try/except IntegrityError` block (the try is a good safety net, but the logic shouldn't be nested inside it).

SdSarthak (changes requested):
The incremental suffix logic (`copy 2`, `copy 3`...) is the right approach, and the `IntegrityError` fallback is a good safety net.

However, there's a critical indentation bug: the new code (`base_name`, `clone_name`, the `while` loop, and `cloned = AISystem(...)`) is at 8-space indent, placing it inside the `if not original: raise HTTPException 404` block. This means the clone creation code will only execute when the system is NOT found (i.e. never on the happy path), and the `try: db.commit()` at 4-space indent will reference an undefined `cloned` variable.

Please fix the indentation so the new code is at 4-space indent (the function body level), matching the original `cloned = AISystem(...)` placement.

SdSarthak:
Closing in favor of PR #1089 which correctly implements the incremental name suffix for clone collisions. This PR had a critical indentation bug where the collision-avoidance code was placed inside the `if not original: raise 404` block, meaning it only ran when the system was not found. PR #1089 has been merged with the correct implementation.

PR #973, opened 2026-06-04
fix: move AI systems filters to server-backed pagination

### LINKED ISSUE #678: bug: Client-side filtering only searches current page, not all records

saurabhhhcodes:
## Summary
- route AI Systems search, risk, and compliance filters through the server-backed list request with the correct `skip` offset
- read the paginated AI Systems response from `items`/`total` so pagination controls reflect the backend result set
- unblock frontend CI by removing unused React default imports, fixing auth error callback typing, allowing document list pagination params, and aligning `ignoreDeprecations` with the installed TypeScript version
- update the changelog

Closes #678

## Validation
- `npm run lint` (frontend)
- `npm run build` (frontend)
- `git diff --check`

## Backend test note
- Attempted `pytest backend/tests/test_ai_systems_filtering.py backend/tests/test_ai_systems_sorting.py` locally with the bundled Python 3.12 runtime. The run is blocked during app import by missing optional runtime packages (`langchain_community`) after resolving earlier local-only dependency mismatches. The targeted backend route already has existing filter/sort tests, and this patch changes the frontend request contract rather than backend filtering logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Unit and integration tests for bypass payloads (`test_normalizer.py` and `test_guard.py`)
 
 ### Fixed
+- **AI Systems** — Send server-side pagination offsets with search and filter requests so results are evaluated across all records instead of the current page only.
 - **Frontend Theme** — Fixed dark mode flash of unstyled content (FOUC), eliminated duplicate CSS, fixed React state overwrite bugs, and improved system preference synchronization.
 - **Documents API** — Validate `ai_system_id` ownership before creating documents so users cannot link documents to another user's AI system.
 - **PDF Export** — Escape user-controlled document text before ReportLab rendering and sanitize generated download filenames.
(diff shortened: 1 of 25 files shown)


saurabhhhcodes:
Closing this as duplicate/stale because #892 has already merged the fix for #678. Keeping the repo queue clean so maintainers do not have to review overlapping PRs.

PR #998, opened 2026-06-10
fix: reject passwords over bcrypt byte limit

saurabhhhcodes:
Closes #932\n\nSummary:\n- reject passwords over bcrypt's 72-byte bcrypt limit in the shared password validator\n- keep registration and change-password paths on the shared validation path\n- add regressions for register and change-password using a UTF-8 password over 72 bytes\n\nValidation:\n- git diff --check\n- /Users/saurabhkumarbajpaiai/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m pytest backend/tests/test_auth.py backend/tests/test_change_password.py -q

### DIFF
--- a/backend/app/api/v1/auth.py
+++ b/backend/app/api/v1/auth.py
@@ -21,7 +21,6 @@ from sqlalchemy.orm import Session
 
 from app.core.database import get_db
 from app.core.security import (
-    validate_password_strength,
     verify_password,
     get_password_hash,
     create_access_token,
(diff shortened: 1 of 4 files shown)


saurabhhhcodes:
This is ready for review. If the repo uses GSSoC scoring labels, please apply the appropriate approval/level label when convenient. The fix is validated with targeted pytest runs for auth and change-password.

PR #833, opened 2026-05-29
Add RAG prompt injection protection, chunk scanning, audit logs, and grounding score

Dhyan-S-Kanapadi:
## Summary

This PR adds production-grade prompt injection protection to the RAG query flow.

It introduces a two-layer guard system:

- Query-level prompt injection scan before retrieval
- Retrieved chunk-level regex scan before context assembly

It also adds RAG audit logging and grounding confidence metadata to the response.

### RAG Query Guard

- Added query-level `LLMGuard.guard()` scan before FAISS retrieval
- Runs guard inference through `run_in_executor` so CPU-bound classifier work does not block the async event loop
- Fails closed if guard inference errors
- Blocks malicious queries before retrieval or LLM execution
- Sanitizes suspicious queries and uses the cleaned prompt downstream
- Logs BLOCK, SANITIZE, ERROR, chunk-drop, and low-grounding events

### Chunk-Level Protection

- Added `LLMGuard.scan_chunk()` for regex-only chunk scanning
- Drops HIGH severity retrieved chunks before context assembly
- Logs warnings for MEDIUM severity chunks
- Returns a safe fallback response when all retrieved chunks are unsafe
- Tracks `chunks_total` and `chunks_dropped`

### Grounding Score

- Added `GroundingChecker`
- Computes cosine similarity between answer embedding and retrieved chunk embeddings
- Returns grounding score and confidence:
  - HIGH: `>= 0.75`
  - MEDIUM: `0.50 - 0.74`
  - LOW: `< 0.50`
- Adds a warning when grounding confidence is LOW

### Audit Logging

- Added `RAGAuditLog` ORM model
- Stores SHA256 question hashes only
- Never stores raw user questions
- Tracks guard decisions, reasoning, chunk drops, grounding score, IP address, and timestamps

### Response Schema

Updated RAG responses to include:

- `grounding_score`
- `grounding_confidence`
- `guard_triggered`
- `guard_decision`
- `chunks_total`
- `chunks_dropped`
- `warning`

### Tests

Added full RAG guard coverage for:

- Benign query with clean chunks
- Query-level BLOCK
- Query-level SANITIZE
- HIGH severity chunk drop
- All retrieved chunks unsafe fallback
- Guard exception fail-closed behavior
- LOW grounding warning
- HIGH grounding without warning

Also stabilized Docker test behavior by isolating test rate limiting from compose Redis.

## Security Notes

- Raw questions are never stored in RAG audit logs
- Audit logs store SHA256 hashes only
[...]

PR #1013, opened 2026-06-12
fix: remove dead rag query response path

saurabhhhcodes:
## What changed

- Removed a dead return path from the AegisAI RAG query flow.

- This keeps the response path simpler and avoids dropping query results unexpectedly.

## Validation

- Not rerun in this session.

### DIFF
--- a/backend/app/api/v1/rag.py
+++ b/backend/app/api/v1/rag.py
@@ -204,7 +204,6 @@ def query_knowledge_base(
             pass
 
         return RAGQueryResponse(answer=answer, sources=sources, answer_id=feedback.id)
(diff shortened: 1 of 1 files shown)


SdSarthak (commented):
Removing the unreachable dead `return` statement after an earlier `return` is a clean fix.

Please rebase onto main to resolve the merge conflict (caused by PR #833 which also modified `rag.py`), then this can be merged.

SdSarthak:
Clean removal of the unreachable dead `return` statement — good catch. This branch is currently in conflict with `main`. Please rebase on the latest main and push again.

PR #1027, opened 2026-06-17
fix(rag): add per-user FAISS index isolation for tenant data protection

### LINKED ISSUE #920: Cross-tenant data poisoning and information disclosure via shared FAISS vector store
### Description

The RAG subsystem stores all user-uploaded documents in a single global FAISS index. The ingestion endpoint rebuilds this index with documents from any authenticated user, and the query endpoint retrieves results from this same shared index. No isolation mechanism restricts document access or modification based on user identity or tenant membership.

### Expected Behavior

Documents uploaded by one user should not be retrievable by other users. The vector store should enforce tenant isolation to prevent data leakage and protect against intentional data poisoning.

### Actual Behavior

**Index creation** (`backend/app/modules/rag/vector_store.py`, lines 35-52):
```python
def load_vector_store():
    # Loads from settings.FAISS_INDEX_PATH
    # No user or tenant parameter
```

**Ingestion** (`backend/app/api/v1/rag.py`, lines 55-131):
```python
@router.post("/ingest")
async def create_vector_store(...):
    # Calls vector_store.create_vector_store() which writes to the global path
    # No user scoping in the index path or document metadata
```

**Query** (`backend/app/modules/rag/retrieval_chain.py`):
```python
# Retrieves from the same global index regardless of querying user
```

No user or tenant identifier is applied anywhere in the chain. The `User` dependency injection is available in the endpoint but is never used to scope data access.

**Attack vectors**:
1. **Data poisoning**: User A uploads PDFs containing deliberately misleading regulatory information. User B's queries may retrieve and incorporate this falsified material into responses.
2. **Information disclosure**: User A uploads proprietary security assessments or internal audit findings. User B's queries may retrieve and surface this confidential information.
3. **Availability degradation**: User A uploads malformed or excessively large documents that corrupt the FAISS index, causing query failures for all users.

**Note**: This issue is distinct from issue #862, which addresses prompt injection persistence through cached retrieval context. That attack vector involves crafted prompts that get cached and returned to other users. This issue concerns the fundamental absence of tenant isolation in the vector store itself, which enables data access and modification across user boundaries regardless of prompt crafting.

### Affected Files

- `backend/app/modules/rag/vector_store.py` (lines 35-52)
- `backend/app/api/v1/rag.py` (lines 55-131)
- `backend/app/modules/rag/retrieval_chain.py`
- `backend/app/core/config.py`

### Steps to Fix

1. Introduce per-user index paths (e.g., `{FAISS_INDEX_PATH}/user_{user_id}/`) to isolate document stores at the filesystem level.
2. Scope `POST /rag/ingest` operations to the calling user's index path.
3. Scope `POST /rag/query` operations to load only the current user's index, with an optional read-only global reference index for common regulatory documents that should be shared.
4. Implement authorization checks in the ingestion and query endpoints to validate user access to indexed data.
5. Add migration logic to split any existing global index into per-user indices.

ionfwsrijan:
## Description

Adds tenant isolation for FAISS vector indexes so that each user's documents are stored in a separate index directory, preventing cross-user data leakage in RAG queries.

### Changes

- **`backend/app/core/config.py`** — Added `FAISS_INDEX_BASE_PATH` setting (default: `./faiss_indices`)
- **`backend/app/modules/rag/vector_store.py`** — Updated `create_vector_store()` and `load_vector_store()` to accept an optional `user_id` parameter. Index files are stored under `{FAISS_INDEX_BASE_PATH}/user_{user_id}/`
- **`backend/app/modules/rag/retrieval_chain.py`** — Threaded `user_id` through the retrieval chain and updated query functions
- **`backend/app/api/v1/rag.py`** — Passes `current_user.id` to RAG endpoints (ingest, query, stream) so indexes are user-scoped

### Security Impact

Previously, all users shared a single FAISS index. With this change, users cannot access or query documents ingested by other users. Existing indexes remain in the old location and continue to work via the default path when `user_id` is `None`.

Closes #920

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Unit and integration tests for bypass payloads (`test_normalizer.py` and `test_guard.py`)
 
 ### Fixed
+- **Per-user FAISS Isolation (#920)** — Added `FAISS_INDEX_BASE_PATH` config and `_get_index_path(user_id)` helper. Vector store functions now accept a `user_id` parameter to store/load indexes under `{FAISS_INDEX_BASE_PATH}/user_{user_id}/`, preventing cross-user data leakage in RAG queries.
 - **Frontend Theme** — Fixed dark mode flash of unstyled content (FOUC), eliminated duplicate CSS, fixed React state overwrite bugs, and improved system preference synchronization.
 - **Documents API** — Validate `ai_system_id` ownership before creating documents so users cannot link documents to another user's AI system.
 - **PDF Export** — Escape user-controlled document text before ReportLab rendering and sanitize generated download filenames.
--- a/backend/app/api/v1/rag.py
+++ b/backend/app/api/v1/rag.py
@@ -82,11 +82,11 @@ def get_rag_guard() -> Any:
     return _RAG_GUARD
 
 
-def get_qa_chain() -> Any:
+def get_qa_chain(user_id: int | None = None) -> Any:
     """Return the configured RAG QA chain."""
     from app.modules.rag.retrieval_chain import get_qa_chain as chain_factory
 
-    return chain_factory()
+    return chain_factory(user_id=user_id)
 
 
 def _hash_question(question: str) -> str:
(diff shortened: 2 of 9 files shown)


ionfwsrijan:
@SdSarthak Please review this now

PR #1030, opened 2026-06-17
fix: resolve pip hash mismatch for smmap and pin compatible versions

### LINKED ISSUE #1029: bug: pip hash mismatch fails Docker build for backend
## Describe the bug
Running `docker compose build` (or `docker compose up -d` on a fresh clone) fails 
during the backend image build with a pip hash verification error.

## Error message:
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have

updated the package versions, please update the hashes. Otherwise, examine the package

contents carefully; someone may have tampered with them.

unknown package:

Expected sha256 5ccb288774fdfb07a7e7025ffec286971c06d8d7b4fb162525334616d7629ff9

Got        5286e04a0babb3e611751fe38f6ff01dea9dafa9810bb1962382d02d37e29a07

## To Reproduce
1. Clone the repo fresh
2. Copy `.env.example` to `.env` and fill in values
3. Run `docker compose up -d`
4. Build fails during `pip install -r requirements.txt` step inside the backend Dockerfile

## Expected behavior
Docker build should complete successfully and all services should start.

## Environment
- OS: Windows 11
- Docker Desktop: 4.73.1
- Docker version: 29.4.3

## Root cause
Several packages in `backend/requirements.txt` are pinned to exact versions 
(e.g. `smmap==5.0.3`) where the locally available wheel hash doesn't match what 
pip expects, likely due to upstream re-releases or transitive dependency conflicts 
with `torch`, `numpy`, `pandas`, `datasets`, and `transformers`.

## Proposed fix
Relax strict version pins (`==`) to minimum version constraints (`>=`) for the 
affected ML/data packages, which resolves the hash conflict while staying within 
compatible version ranges. PR incoming with the fix.

## Screenshot 
Image

VaishnaviJagtap18:
## Summary

Closes #1029

This PR fixes a pip hash mismatch error that occurs when building the backend Docker image. Several packages (smmap, torch, numpy, pandas, datasets, transformers, scikit-learn) had strict version pins causing "THESE PACKAGES DO NOT MATCH THE HASHES" errors during `docker compose build`. Relaxed pins to minimum version constraints to resolve this.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor
- [ ] Tests
- [ ] Infra / CI

## Checklist

## Before

error- setup

## After

built

error fixed

### DIFF
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -11,7 +11,7 @@ alembic==1.13.1
 # ── Auth ──────────────────────────────────────────────────────────────────────
 python-jose[cryptography]==3.3.0
 passlib[bcrypt]==1.7.4
-bcrypt==3.2.2
+bcrypt==4.0.1
 
 # ── Settings & Validation ─────────────────────────────────────────────────────
 pydantic==2.5.3


SdSarthak (changes requested):
Thanks for addressing the pip hash mismatch. The bcrypt upgrade from 3.2.2 → 4.0.1 is correct and needed.

However, there are some concerns with the broader changes:

1. **Removing `faiss-cpu`** — the RAG module actively uses FAISS for vector search. Removing it would break `POST /rag/ingest` and `POST /rag/query` at startup. Please restore it.
2. **Removing `mlflow` and `boto3`** — these may be optional for some deployments but removing them without confirming all import paths is risky. If they're truly unused, that's a separate PR.
3. **Switching to loose `>=` constraints** for ML dependencies (torch, transformers, etc.) — version ranges this wide can cause compatibility issues between packages. Prefer tight pins or minor-version ranges (e.g. `torch>=2.1,<3`).
4. **Adding `smmap>=5.0.0`** — `smmap` is a git internal dependency and shouldn't be in `requirements.txt`.

Please limit this PR to just the bcrypt version upgrade and rebase on main.

VaishnaviJagtap18:
Thanks for the detailed review @SdSarthak. Narrowed the PR to just the bcrypt 3.2.2 → 4.0.1 upgrade. All other packages restored to their original pins. Rebased on main.

VaishnaviJagtap18:
Hi @SdSarthak, just checking in, I've already pushed the updated commit (55784bb) with only the bcrypt 3.2.2 → 4.0.1 change, all other packages restored to original pins, and rebased on main. Could you please re-review when you get a chance? Happy to make any further changes if needed!

PR #1113, opened 2026-06-22
fix : added empty file and zero-byte validation for RAG document ingestion

### LINKED ISSUE #1108: fix : add empty file and zero-byte validation for RAG document ingestion
load_documents_from_paths in backend/app/modules/rag/document_loader.py accepts any file path without validating that the file is non-empty or above a minimum byte threshold. Zero-byte or effectively empty PDFs (e.g. corrupted or placeholder files) pass through to PyPDFLoader and then to the FAISS index builder, polluting the vector store with empty chunks. The endpoint in backend/app/api/v1/rag.py does check file_size > RAG_MAX_FILE_SIZE_BYTES but does not check for size == 0. The document_loader itself has no guard.

Changes that Need to be Made:
- backend/app/modules/rag/document_loader.py: Add a pre-flight check in load_documents_from_paths that skips files smaller than a configurable minimum (e.g. 100 bytes) and raises a ValueError with a descriptive message listing any rejected paths. Also add a try/except around each PyPDFLoader.load() call so a single corrupt file does not abort the entire batch.

Impact that it would Provide:
- Prevents empty or placeholder PDFs from polluting the FAISS index
- Provides a clear error message for users instead of a silent empty-chunk index
- Hardens the ingestion pipeline against corrupted file uploads

tmdeveloper007:
Closes #1108.

Added a minimum file size check (_MIN_PDF_SIZE_BYTES = 100 bytes) in load_documents_from_paths before passing paths to PyPDFLoader. Empty and zero-byte PDFs now raise a descriptive ValueError listing the rejected paths. Also wrapped each PyPDFLoader.load() call in try/except so a single corrupt PDF aborts with a clear error message instead of a raw traceback.

Changes Made:
- backend/app/modules/rag/document_loader.py: Added _MIN_PDF_SIZE_BYTES constant. load_documents_from_paths now checks each path's size first and collects all rejected paths before raising a ValueError with the full list. Each loader.load() call is wrapped in try/except with a descriptive ValueError on failure.

Impact it Made:
- Prevents empty or placeholder PDFs from polluting the FAISS index with empty chunks
- Provides clear, actionable error messages for users uploading invalid files
- Hardens the ingestion pipeline against corrupted uploads

### DIFF
--- a/backend/app/modules/rag/document_loader.py
+++ b/backend/app/modules/rag/document_loader.py
@@ -1,9 +1,17 @@
 """Document loader for ingesting regulatory PDFs from S3 or local disk."""
 
+import logging
 import os
-from langchain_community.document_loaders import S3DirectoryLoader, PyPDFLoader
+
+from langchain_community.document_loaders import PyPDFLoader, S3DirectoryLoader
 from langchain_text_splitters import RecursiveCharacterTextSplitter
+
 from app.core.config import settings
+
+logger = logging.getLogger(__name__)
+
+# Minimum file size in bytes — anything below this is considered an empty or placeholder file.
+_MIN_PDF_SIZE_BYTES = 100
 
 
 def load_documents_from_s3():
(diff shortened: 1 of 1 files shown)


zycaskevin:
## Code Review: Great direction, a few improvements I'd suggest 👍

The core idea is spot-on — empty/corrupt PDFs polluting the FAISS index is a real problem and fixing it at the `document_loader` layer is the right place.

### ✅ What's good
- Collecting **all** rejected paths before raising (not fail-fast on first) — very user-friendly
- Wrapping `PyPDFLoader.load()` in try/except prevents one corrupt file from aborting the whole batch
- Clear error message with the file list

### 🔧 Suggested improvements

#### 1. Make the threshold configurable (not a magic constant)

Hard-coded constants make it hard to tune for different environments (dev vs prod). Making it a `Settings` field is discoverable and configurable via `RAG_MIN_FILE_SIZE_BYTES` env var.

#### 2. Use specific exception instead of bare `except Exception`

A bare `except Exception` could swallow `PermissionError`, `OSError`, or other unexpected errors. Use `pypdf.errors.PdfReadError` (or `pypdf.PdfReadError` depending on version).

#### 3. Add 0-byte check at the API endpoint
The `document_loader` handles files >= min_size but doesn't reject **exactly** 0 bytes differently from "near-zero". The API endpoint in `rag.py` checks `RAG_MAX_FILE_SIZE_BYTES` but has no 0-byte check. Adding it at the API layer gives users a fast 400 error before any disk write:


#### 4. Handle OSError from `os.path.getsize()`
If a file is deleted between listing and stat, or permission is denied:


#### 5. Add tests ⚠️
The original PR has **zero tests**. This is the most important gap. Suggested test cases:
- Empty (0-byte) file → raises `ValueError`
- File below min_size → raises `ValueError`
- Corrupt PDF (passes size check) → raises `ValueError` with log warning
- Multiple empty files → lists ALL in one error
- Valid PDF → loads normally (mocked `PyPDFLoader`)
- Configurable threshold → works when `RAG_MIN_FILE_SIZE_BYTES` is changed

---

### PR #1118 (improvement branch)
I opened a follow-up PR [#1118](https://github.com/SdSarthak/AegisAI/pull/1118) that applies all of the above suggestions. It includes:
- Configurable `RAG_MIN_FILE_SIZE_BYTES` in settings
- Specific `pypdf.errors.PdfReadError` handling
- API-layer 0-byte check
- 9 comprehensive tests (all passing)

Would be great if the original author could review it or cherry-pick ideas. Great start! 🚀

PR #1115, opened 2026-06-22
fix : added questionnaire_responses key/value whitespace trimming in compliance engine

### LINKED ISSUE #1110: fix : trim and sanitize compliance engine questionnaire_responses keys and values
The evaluate_compliance function in backend/app/modules/compliance/eu_ai_act.py passes questionnaire_responses directly into the requirement status lookup without normalizing the keys or values. If a client sends keys with leading/trailing whitespace (e.g. ' article_9 ' instead of 'article_9') or string values with extra spaces (e.g. ' done ' instead of 'done'), the key lookup fails and status defaults to 'missing', incorrectly marking requirements as non-compliant.

Changes that Need to be Made:
- backend/app/modules/compliance/eu_ai_act.py: In evaluate_compliance, strip whitespace from each key in questionnaire_responses before lookup (key.strip()). Also strip and lowercase the req_status value before validating against ('missing', 'partial', 'done').

Impact that it would Provide:
- Prevents false 'missing' compliance statuses caused by accidental whitespace in API payloads
- Makes the compliance engine more robust against real-world client variations
- Ensures compliance assessment results are accurate regardless of input formatting

tmdeveloper007:
Closes #1110.

evaluate_compliance in eu_ai_act.py now strips whitespace from all questionnaire_responses keys and values before lookup and status evaluation. A key of ' article_9 ' is now correctly matched as 'article_9', and a status value of ' done ' is canonicalized to 'done' instead of falling through to 'missing'.

Changes Made:
- backend/app/modules/compliance/eu_ai_act.py: Added _normalize_status helper that strips and lowercases status values and maps legacy labels via _COMPLIANCE_STATUS_MAP. Added normalized_responses dict that strips all keys and normalizes all values upfront. Replaced direct get() with the normalized view.

Impact it Made:
- Prevents false 'missing' compliance statuses from accidental whitespace in client payloads
- Makes compliance assessment results accurate regardless of input formatting

### DIFF
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -1,7 +1,10 @@
 """Shared pytest fixtures for all tests."""
 
+from __future__ import annotations
+from tests.csrf_client import _CSRFClientWrapper
 import os
 import pytest
+from typing import Any
 from unittest.mock import MagicMock
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker, Session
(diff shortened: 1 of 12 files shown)


SdSarthak:
This PR conflicts with recently-merged changes to `backend/app/modules/compliance/eu_ai_act.py`. Please rebase on `main` and resolve the conflict — the whitespace-trimming logic in this PR is solid and worth keeping.

Issue #1106, opened 2026-06-21
feat: Add empty file and size validation for RAG document ingestion


DeepeshKafalatiya:
### What problem does this solve? Why is it important?

Currently, the RAG Intelligence ingestion endpoint allows users to upload source files (like PDFs or TXT files containing compliance rules). However, if an uploaded file is empty (0 bytes) or corrupted, the system still passes it down to the document loaders (like `PyPDFLoader`). 

This results in downstream errors, such as:
1. `PyPDFLoader` throwing unhandled processing exceptions.
2. The text splitter producing empty chunks.
3. Wasteful embedding API calls or unindexable vector insertions in the FAISS store.

Catching empty files or establishing a sensible maximum file size check directly at the endpoint level improves API resilience and prevents corrupt data from contaminating the vector store.

### Proposed Solution

1. Locate the document upload router file for the RAG engine.
2. In the FastAPI route handling the `UploadFile` stream, read the file size or read a small chunk to ensure it contains content:
```python
   # Example check
   if file.size == 0:
       raise HTTPException(status_code=400, detail="Uploaded file is empty.")
Add a soft validation check ensuring that the file does not exceed a reasonable maximum threshold (e.g., 20MB) to prevent server-side memory bottlenecks during chunk extraction.

Return a clean HTTP 400 Bad Request error with a descriptive message if validation fails.

Files to look at:

backend/app/api/v1/endpoints/rag.py (or the core RAG router module)

### Which module does this touch?

RAG Intelligence

### Alternatives Considered

Handling validations on the client-side, but relying solely on frontend constraints allows direct API calls to bypass checks and potentially stress the text processing pipeline with broken documents.

DeepeshKafalatiya:
Hi @SdSarthak  👋

I noticed that adding quick content validations on the RAG document ingestion routes would prevent the vector database from indexing empty text chunks and stop file processors from throwing raw runtime errors.

I'm participating in GSSoC and would love to work on implementing this file validation safely. Could you please assign this issue to me? Thank you!

ujala100:
Hi @SdSarthak

I’m interested in contributing to this project and would love to work on this issue. I have experience working on open-source contributions and would be happy to implement the required changes.

Could you please assign this issue to me? Thanks

Nakshathra-2808:
### Proposed Fix Approach

[...]

PR #1118, opened 2026-06-22
fix(rag): harden document ingestion validation & tests

### LINKED ISSUE #1108: fix : add empty file and zero-byte validation for RAG document ingestion

zycaskevin:
## Summary

This PR hardens the RAG document ingestion pipeline to prevent corrupted, empty, or placeholder PDFs from polluting the FAISS vector store. It improves upon the initial approach in PR #1113 (closes #1108 / #1106).

### `config.py`
- **Added `RAG_MIN_FILE_SIZE_BYTES`** (default 100 bytes) — makes the threshold configurable via environment instead of a magic number

### `document_loader.py`
- **Pre-flight size check**: rejects files below `RAG_MIN_FILE_SIZE_BYTES` with a descriptive `ValueError` listing ALL rejected paths
- **OSError handling**: files that fail `os.path.getsize` are rejected with the error message
- **Specific exception handling**: `pypdf.errors.PdfReadError` is caught separately with a warning log, instead of a bare `except Exception`
- **Better error messages**: includes the filename in the `ValueError` for quick debugging

### `rag.py` (API endpoint)
- **0-byte rejection**: uploads with `file_size == 0` are rejected immediately with a clear 400 message before any disk write

### `tests/test_document_loader.py` (NEW)
- 9 comprehensive tests covering:
  - Empty (0-byte) file rejection
  - Below-minimum-size rejection
  - Configurable threshold via settings
  - Multiple empty files listed in single error
  - Corrupt/invalid PDF rejection
  - Warning log emission for corrupt files
  - OSError/unreadable file rejection
  - Valid PDF processing (mocked)
  - Corrupt file in a batch raises correctly

## Test Results
```
tests/test_document_loader.py ............ 9 passed
```

## Why not just hard-code 100 bytes?
The original PR #1113 used a magic constant `_MIN_PDF_SIZE_BYTES = 100`. This makes it:
- Non-configurable for different environments (dev vs prod may have different needs)
- Harder to tune if the default turns out to be too aggressive or too lenient
- Not discoverable via `settings` introspection

Making it a Settings field solves all three.

## Comparison with original PR #1113

| Aspect | #1113 | This PR |
|--------|-------|---------|
| Size validation | ✅ 100 bytes hard-coded | ✅ Configurable `RAG_MIN_FILE_SIZE_BYTES` |
| Exception handling | `except Exception` | `except PdfReadError` (specific) + logging |
| API 0-byte check | ❌ | ✅ 400 error before disk write |
| Tests | ❌ | ✅ 9 tests |
| OSError handling | ❌ | ✅ Graceful error message |

### DIFF
--- a/backend/app/api/v1/rag.py
+++ b/backend/app/api/v1/rag.py
@@ -321,6 +321,15 @@ def ingest_documents(
         file_size = upload.file.tell()
         upload.file.seek(0)
 
+        if file_size == 0:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=(
+                    f"File {upload.filename} is empty (0 bytes). "
+                    "Please upload a valid PDF file."
+                ),
+            )
+
         if file_size > settings.RAG_MAX_FILE_SIZE_BYTES:
             raise HTTPException(
                 status_code=status.HTTP_413_REQUEST_ENTITY_TOO_LARGE,
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -62,6 +62,7 @@ class Settings(BaseSettings):
     MLFLOW_TRACKING_URI: str = ""
     EMBEDDINGS_MODEL: str = "nomic-embed-text"
     RAG_MAX_FILES_PER_REQUEST: int = 10
+    RAG_MIN_FILE_SIZE_BYTES: int = 100  # Minimum valid PDF size in bytes
     RAG_MAX_FILE_SIZE_BYTES: int = 10 * 1024 * 1024
     RAG_TOTAL_BUDGET_BYTES: int = 50 * 1024 * 1024
 
--- a/backend/app/modules/rag/document_loader.py
+++ b/backend/app/modules/rag/document_loader.py
@@ -1,9 +1,15 @@
 """Document loader for ingesting regulatory PDFs from S3 or local disk."""
 
+import logging
 import os
+import pypdf.errors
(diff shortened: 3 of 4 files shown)


SdSarthak:
This PR conflicts with recently-merged changes to `backend/app/modules/rag/document_loader.py`. Please rebase on `main` and resolve the conflict — the validation and error handling improvements here are comprehensive and worth keeping.

PR #1132, opened 2026-06-22
fix : added cleanup_stale_local_attempts on every _check_local call to prevent unbounded memory growth

### LINKED ISSUE #1127: fix : call cleanup_stale_local_attempts on every _check_local call to prevent unbounded memory growth
Summary: The DistributedRateLimiter._check_local method only calls _cleanup_stale_local_attempts_locked when the rate limit is exceeded. Active-but-not-limited keys with expired entries are never cleaned up, causing unbounded memory growth.

Changes: Move the cleanup call outside the if len(attempts) + cost > limit block so it runs on every _check_local invocation. This ensures old entries are always pruned regardless of whether the key is limited.

Impact: prevents unbounded memory growth in the in-memory rate limiter fallback; cleaner memory usage in long-running processes.

Note: Please assign this issue to the tmdeveloper007 account.

tmdeveloper007:
Closes #1127.

Moved the periodic cleanup counter increment and stale-entry pruning (cleanup_stale_local_attempts_locked) to run on every _check_local invocation, before the rate-limit check. Previously cleanup only ran on non-rate-limited calls; rate-limited calls returned early without cleaning up, causing unbounded memory growth for keys that are always limited.

Changes Made:
- backend/app/core/rate_limit.py: moved _local_cleanup_calls increment and conditional cleanup to run before the rate-limit check, ensuring cleanup runs on every call

Impact it Made:
- Active-but-not-limited keys no longer accumulate unbounded memory
- Rate-limited keys are also cleaned up on every invocation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,4 +119,4 @@ pdf_export.txt
 rag_ingest.txt
 
 # Models
-backend/app/modules/guard/models/*
+backend/app/modules/guard/models/*backend/rag_documents/
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -8,6 +8,7 @@ from sqlalchemy.orm import sessionmaker, Session
 from sqlalchemy.pool import StaticPool
 from fastapi import Request, HTTPException, status
 from fastapi.testclient import TestClient
+from starlette.responses import Response
 
 # Set test database before importing app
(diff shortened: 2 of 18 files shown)


SdSarthak:
This PR conflicts with recently-merged changes to `backend/tests/conftest.py`. Please rebase on `main` and resolve the conflict.

PR #1174, opened 2026-06-23
fix : moved registration rate-limit record out of finally block in auth.py

### LINKED ISSUE #1169: fix : move registration rate-limit record out of finally block in auth.py
In backend/app/api/v1/auth.py:195-200, POST /register calls _record_attempt() inside a finally block. This means even successful registrations consume the rate-limit budget. After 3 successful registrations from one IP, the 4th valid registration is incorrectly rejected with HTTP 429. The login endpoint correctly records attempts only on failure (auth.py:238).

Changes that Need to be Made:
- backend/app/api/v1/auth.py: Move _record_attempt() out of the finally block into the failure branches (except HTTPException and the generic Exception handler), following the pattern used in the login endpoint.

Impact that it would Provide:
- Rate limiting for registration will only count actual failures, not successful sign-ups
- Prevents false positive 429 rejections for legitimate users behind shared IPs
- Consistent behavior with the login rate limiter

tmdeveloper007:
Closes #1169.

Moved _record_attempt() call out of the finally block in POST /register into the failure branches (duplicate email, server error). Successful registrations no longer consume the rate-limit budget. Updated the rate-limit test to use duplicate email for repeated failures.

Changes Made:
- backend/app/api/v1/auth.py: Moved _record_attempt from finally to the duplicate-email and server-error branches
- backend/tests/test_auth.py: Updated test_register_rate_limit_triggers_after_three_attempts to use duplicate email

Impact it Made:
- Successful registrations no longer cause false-positive 429 rejections
- Consistent with login endpoint behavior (only failures count)
- All 17 auth tests pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,4 +119,4 @@ pdf_export.txt
 rag_ingest.txt
 
 # Models
-backend/app/modules/guard/models/*
+backend/app/modules/guard/models/*backend/rag_documents/
--- a/backend/app/api/v1/auth.py
+++ b/backend/app/api/v1/auth.py
@@ -162,6 +162,11 @@ def register(
     try:
         existing_user = db.query(User).filter(User.email == user_data.email).first()
         if existing_user:
+            _record_attempt(
+                _auth_registration_attempts_by_ip,
+                client_ip,
+                _AUTH_REGISTER_RATE_LIMIT_WINDOW_SECONDS,
+            )
             raise HTTPException(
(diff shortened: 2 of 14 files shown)


SdSarthak:
This PR conflicts with recently-merged changes to `backend/tests/conftest.py` (via #1165). Please rebase on `main` and re-push — the core change here is good and will be merged once the conflict is resolved.

PR #1201, opened 2026-06-24
fix : use useRef for saveTimeout to prevent infinite re-render in DocumentEditor

### LINKED ISSUE #1196: fix : use useRef for saveTimeout to prevent infinite re-render in DocumentEditor
In DocumentEditor.tsx, saveTimeout is stored in useState and included in the useEffect dependency array, creating an infinite re-render loop. Every time the user types, the effect fires, calls setSaveTimeout, which triggers a re-render, which fires the effect again, immediately clearing the timeout before it can fire. Auto-save never executes.

Changes that Need to be Made:
- frontend/src/components/DocumentEditor.tsx: Replace useState for saveTimeout with useRef. Store the timeout ID in saveTimeoutRef.current instead of React state. Remove saveTimeout from the useEffect dependency array.

Impact that it would Provide:
- Auto-save fires correctly 2 seconds after the user stops typing
- Eliminates the infinite re-render loop
- Better performance and user experience

tmdeveloper007:
Closes #1196.

Replaced useState for saveTimeout with useRef in DocumentEditor.tsx. The saveTimeout was stored in React state and included in the useEffect dependency array, creating an infinite re-render loop where every state update triggered the effect again, immediately clearing the timeout before it could fire.

Changes Made:
- frontend/src/components/DocumentEditor.tsx: Changed saveTimeout from useState to useRef, updated clearTimeout call to use saveTimeoutRef.current, and removed saveTimeout from the useEffect dependency array.

Impact it Made:
- Auto-save now fires correctly 2 seconds after user stops typing
- Eliminates the infinite re-render loop that was causing performance issues
- Better user experience

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,4 +119,4 @@ pdf_export.txt
 rag_ingest.txt
 
 # Models
-backend/app/modules/guard/models/*
+backend/app/modules/guard/models/*backend/rag_documents/
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -8,6 +8,7 @@ from sqlalchemy.orm import sessionmaker, Session
 from sqlalchemy.pool import StaticPool
 from fastapi import Request, HTTPException, status
 from fastapi.testclient import TestClient
+from starlette.responses import Response
 
 # Set test database before importing app
 os.environ["DATABASE_URL"] = "sqlite:///:memory:"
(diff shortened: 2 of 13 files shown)


SdSarthak:
This PR conflicts with recently-merged changes to `backend/tests/conftest.py` (via #1165). Please rebase on `main` and re-push — the code change here is correct and will be merged once conflicts are resolved.

Issue #1170, opened 2026-06-23
fix : replace datetime.utcnow() with datetime.now(timezone.utc) in guard.py


tmdeveloper007:
Summary of What Needs to be Done:
GuardScanLog.scanned_at uses naive datetime.utcnow() (backend/app/api/v1/guard.py:161, 766), but cursor pagination encoding/decoding makes datetimes timezone-aware (lines 492-510). On PostgreSQL, comparing naive (database column) with aware (query parameter) causes: 'cannot compare timestamp without time zone with timestamp with time zone'. On SQLite the comparison silently produces wrong results.

Changes that Need to be Made:
- backend/app/api/v1/guard.py: Replace all datetime.utcnow() with datetime.now(timezone.utc). Affected lines: 161 (GuardScanLog creation), 766 (stats endpoint). Import timezone from datetime.

Impact that it would Provide:
- Guard history date filtering will work correctly on PostgreSQL without runtime crashes
- Consistent timezone-aware storage across all datetime fields
- Fixes silent incorrect results on SQLite

Note: Please assign this issue to the `tmdeveloper007` account.

PR #1212, opened 2026-06-25
fix : replaced datetime.utcnow() with datetime.now(timezone.utc) in GuardScanLog

### LINKED ISSUE #1207: fix : replace datetime.utcnow() with datetime.now(timezone.utc) in GuardScanLog
Replace naive datetime.utcnow() with timezone-aware datetime.now(timezone.utc) in the GuardScanLog model and the scan endpoint. GuardScanLog.scanned_at uses datetime.utcnow (naive) but cursor pagination decode always produces timezone-aware datetimes, causing PostgreSQL to crash when comparing naive vs aware datetimes in the /guard/history endpoint.

Changes that Need to be Made:
- backend/app/models/guard_scan_log.py: Change scanned_at default from datetime.utcnow to datetime.now(timezone.utc) and add timezone import
- backend/app/api/v1/guard.py: Change scanned_at=datetime.utcnow() to scanned_at=datetime.now(timezone.utc) in the _build_guard_scan_log function (line 161)

Impact that it would Provide:
- Fixes PostgreSQL runtime crash when using cursor pagination on /guard/history with start_date or end_date filters
- Consistent timezone-aware timestamps throughout the codebase
- Follows the same fix pattern as the already-merged PR #1170

tmdeveloper007:
Closes #1207.

Replaced naive datetime.utcnow() with timezone-aware datetime.now(timezone.utc) in the GuardScanLog model (scanned_at and created_at defaults) and in the _build_guard_scan_log function in guard.py. This fixes the PostgreSQL crash when cursor pagination compares a timezone-aware datetime (produced by cursor decode) against the naive database column.

Changes Made:
- backend/app/models/guard_scan_log.py: Changed datetime import to include timezone; updated scanned_at and created_at defaults to use datetime.now(timezone.utc)
- backend/app/api/v1/guard.py: Updated scanned_at=datetime.utcnow() to scanned_at=datetime.now(timezone.utc)

Impact it Made:
- Fixes PostgreSQL 'cannot compare timestamp without time zone with timestamp with time zone' crash
- Consistent timezone-aware timestamps in GuardScanLog
- Follows the same pattern as the already-merged PR #1170 (datetime fix in guard.py)

### DIFF
--- a/backend/app/api/v1/guard.py
+++ b/backend/app/api/v1/guard.py
@@ -158,7 +158,7 @@ def _build_guard_scan_log(user_id: int, prompt: str, result: dict, ip_address: s
         ml_confidence=intent_analysis.get("confidence", 0.0),
         combined_score=decision_reasoning.get("confidence", 0.0),
         prompt_length=len(prompt),
-        scanned_at=datetime.utcnow(),
+        scanned_at=datetime.now(timezone.utc),
         ip_address=ip_address,
     )
 
--- a/backend/app/models/guard_scan_log.py
+++ b/backend/app/models/guard_scan_log.py
@@ -6,7 +6,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 Prompts are stored as SHA-256 hashes only — never raw text — to protect user privacy.
 """
 
-from datetime import datetime
+from datetime import datetime, timezone
 
 from sqlalchemy import Column, Integer, String, Float, DateTime, ForeignKey, JSON, Boolean
 from sqlalchemy.orm import relationship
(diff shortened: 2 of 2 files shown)


SdSarthak:
Thanks for the update, but this still doesn't address the review comment — `scanned_at` and `created_at` still use `default=datetime.now(timezone.utc)` (a plain value computed once when the class body executes), not a callable. Every row will get the same frozen timestamp from whenever the module was imported. Please change both to `default=lambda: datetime.now(timezone.utc)` so the value is computed per-row at insert time.

PR #1225, opened 2026-06-26
fix : prevent GuardScanLog IntegrityError and fix conftest.py TestClient.response

### LINKED ISSUE #1220: bug : GuardScanLog silently lost when create_notification fails and rollback re-attaches detached object
In log_scan() (backend/app/api/v1/guard.py), when create_notification() fails and db.rollback() is called, the code attempts to re-add the log object with db.add(log). However, log was already committed at an earlier step, so after rollback it is in a detached state. Calling db.add() on a detached object causes IntegrityError on the next db.commit(), which is silently caught by the outer except block - losing the entire scan log entry silently.

Changes that Need to be Made:
- backend/app/api/v1/guard.py: Either (a) move db.commit() after create_notification succeeds, or (b) check if log is attached before re-adding after rollback, or (c) use a savepoint/nested transaction so rollback only affects the notification step. Option (a) is simplest: commit log only after create_notification succeeds.

Impact that it would Provide:
- Guard scan logs are never silently lost
- Consistent audit trail for compliance requirements
- IntegrityError becomes visible and actionable

tmdeveloper007:
Closes #1220.

Fixed two related issues in backend/app/api/v1/guard.py log_scan():
1. Removed the redundant db.add(log) + db.commit() after db.rollback() in the create_notification error handler. The log entry was already committed before create_notification was called, so re-committing it after rollback could cause IntegrityError.
2. Cherry-picked the conftest.py TestClient.response -> Response fix which was blocking pytest from importing the test suite.

Changes Made:
- backend/app/api/v1/guard.py: Removed redundant db.add(log) + db.commit() after rollback; the committed log is safe and does not need re-committing.
- backend/tests/conftest.py: Changed _CSRFClientWrapper return types from TestClient.response to starlette.responses.Response.

Impact it Made:
- Guard scan logs are no longer silently lost when create_notification fails
- conftest.py pytest import error is resolved

### DIFF
--- a/backend/app/api/v1/guard.py
+++ b/backend/app/api/v1/guard.py
@@ -189,8 +189,6 @@ def log_scan(user_id: int, prompt: str, result: dict, ip_address: str | None = N
             except Exception:
                 db.rollback()
                 logger.warning("Failed to create block notification for scan %d", log.id)
-                db.add(log)
-                db.commit()
 
     except Exception:
         db.rollback()
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -8,6 +8,7 @@ from sqlalchemy.orm import sessionmaker, Session
 from sqlalchemy.pool import StaticPool
 from fastapi import Request, HTTPException, status
 from fastapi.testclient import TestClient
+from starlette.responses import Response
 
 # Set test database before importing app
 os.environ["DATABASE_URL"] = "sqlite:///:memory:"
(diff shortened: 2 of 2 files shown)


SdSarthak:
This PR conflicts with recently-merged changes to `backend/tests/conftest.py` (via #1165). Please rebase on `main` and re-push — the changes here look good and will be reviewed for merge once the conflict is resolved.

PR #1232, opened 2026-06-26
Fix analytics summary loading

### LINKED ISSUE #1230: Analytics page ignores shared API client and shows hardcoded/fallback data
### Describe the bug

# Bug: Analytics page ignores shared API client and shows hardcoded/fallback data

### Summary

The Analytics page fetches `/api/v1/analytics/summary` directly and only uses the API response for the risk pie chart. Other visible analytics cards and charts are still hardcoded. In deployments with a custom `VITE_API_BASE_URL`, the direct fetch can hit the wrong host/path, fail, and silently display fallback data.

### Suggested fix

- Replace the direct `fetch()` call with `analyticsApi.summary()`.
- Map `total_systems`, `average_compliance_score`, `counts`, and `compliance_statuses` into the visible summary cards and charts.
- Show an error/empty state instead of silently replacing failed API data with fake values.

### Impact

- Analytics works in custom deployment paths.
- Users see accurate compliance and risk metrics.
- Avoids misleading dashboards that look successful while showing placeholder values.

### Steps to reproduce

1. Configure the frontend with `VITE_API_BASE_URL` pointing to a non-root backend URL.
2. Open the Analytics page.
3. The page still calls `/api/v1/analytics/summary` directly instead of the configured API base.
4. If the call fails, the page silently shows fallback risk data and hardcoded summary cards.

### Expected behaviour

Analytics should use the configured shared API client and render live backend summary data wherever available.

### Actual behaviour

The page bypasses the shared API client, ignores `VITE_API_BASE_URL`, and can show hardcoded or fallback numbers without clearly surfacing the failed API call.

### AegisAI version / commit

_No response_

Tomeshwari-02:
## Summary

- load analytics data through the shared API client
- render summary cards and risk charts from the backend summary response
- show an error state instead of replacing failed requests with fallback data

## Root cause

The Analytics page was calling `/api/v1/analytics/summary` directly and still rendered hardcoded/fallback values. That bypassed the configured API base URL and could show misleading analytics when the request failed.

## Validation

- `npm run build`

Closes #1230

### DIFF
--- a/frontend/src/pages/Analytics.tsx
+++ b/frontend/src/pages/Analytics.tsx
@@ -1,5 +1,6 @@
 import { useEffect, useState } from "react";
 import { formatLastUpdated } from "../utils/date";
+import { analyticsApi } from "../services/api";
 
 import ComplianceRiskChart from "../components/ComplianceRiskChart";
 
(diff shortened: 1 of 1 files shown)


SdSarthak:
This PR conflicts with recently-merged changes. Please rebase on `main` — the analytics summary loading fix with real API calls is a solid improvement and will be merged once the conflict is resolved.

PR #1237, opened 2026-06-26
Fix auth revalidation API base

### LINKED ISSUE #1235: Auth session revalidation hardcodes `/api/v1/auth/me` and ignores configured API base URL
### Describe the bug

# Bug: Auth session revalidation hardcodes `/api/v1/auth/me` and ignores configured API base URL

### Summary

`authStore.revalidateSession()` calls `fetch('/api/v1/auth/me')` directly instead of using the shared API client or the configured API base URL. In deployments where the backend is not served from the frontend origin at `/api/v1`, session revalidation fails and the store logs the user out.

### Code evidence

- `frontend/src/stores/authStore.ts:33` defines `revalidateSession`
- `frontend/src/stores/authStore.ts:41` calls `fetch('/api/v1/auth/me', ...)`
- `frontend/src/stores/authStore.ts:54` logs the user out on network error
- `frontend/src/services/api.ts:5` centralizes `VITE_API_BASE_URL`
- `frontend/src/services/api.ts:150` already provides `authApi.getMe()`

### Suggested fix

- Refactor revalidation to use the shared API URL builder or `authApi.getMe(token)`.
- Avoid circular imports between `api.ts` and `authStore.ts`; if needed, move `buildApiUrl()` into a small shared utility.
- Consider distinguishing network errors from confirmed `401`/`403` token failures before clearing auth state.

### Impact

- Prevents users from being logged out after refresh in custom deployments.
- Keeps auth behavior consistent with `VITE_API_BASE_URL`.
- Reduces false session loss caused by network/path configuration issues.

### Steps to reproduce

1. Deploy the frontend with `VITE_API_BASE_URL` pointing to a backend on another origin or under a non-root path.
2. Log in successfully through the configured API client.
3. Refresh the app or trigger session revalidation.
4. `revalidateSession()` calls `/api/v1/auth/me` on the frontend origin instead of `VITE_API_BASE_URL`.
5. The request fails and the user is logged out.

### Expected behaviour

Session revalidation should use the same configured API base URL as the rest of the frontend API calls.

### Actual behaviour

Revalidation bypasses the shared API client and can incorrectly clear a valid session.

### AegisAI version / commit

_No response_

Tomeshwari-02:
## Summary

- move API base URL handling into a shared frontend utility
- use the shared URL builder for auth session revalidation
- keep the persisted session intact when revalidation fails because of a transient network error

## Root cause

Session revalidation called `/api/v1/auth/me` directly from the auth store. That bypassed `VITE_API_BASE_URL`, so deployments where the API lives on a different origin or base path could incorrectly clear a valid session.

## Validation

- `npm run build`

Closes #1235

### DIFF
--- a/frontend/src/services/api.ts
+++ b/frontend/src/services/api.ts
@@ -1,16 +1,9 @@
 import axios, { InternalAxiosRequestConfig, AxiosResponse } from 'axios'
 import { useAuthStore } from '../stores/authStore'
-
-const configuredApiBaseUrl = import.meta.env.VITE_API_BASE_URL?.trim()
-const API_BASE_URL = configuredApiBaseUrl ? configuredApiBaseUrl.replace(/\/$/, '') : '/api/v1'
+import { API_BASE_URL, buildApiUrl } from '../utils/apiUrl'
 
 // Tracks whether the global 401-response handler is currently executing.
 let isUnauthorizedHandlerRunning = false
-
-function buildApiUrl(path: string): string {
-  const normalizedPath = path.startsWith('/') ? path : `/${path}`
-  return `${API_BASE_URL}${normalizedPath}`
-}
 
 const api = axios.create({
   baseURL: API_BASE_URL,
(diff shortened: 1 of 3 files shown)


SdSarthak:
The 2026-07-03 review flagged a security regression: the catch block sets isAuthenticated: true on network error, which keeps an unverifiable session alive indefinitely. No commits have been pushed since that review, so this is still unresolved. Please update the catch block to leave the auth state unchanged on network error instead of asserting it true.

PR #1241, opened 2026-06-26
fix : move registration rate-limit record out of finally block in auth.py

### LINKED ISSUE #1169: fix : move registration rate-limit record out of finally block in auth.py

tmdeveloper007:
Closes #1169.

Moved _record_attempt() calls out of the register() finally block and into the exception handlers. This ensures only failed registration attempts (duplicate email, database errors) count toward the rate limit, not successful registrations. Also updated the rate-limit test to reflect the corrected behavior.

Changes Made:
- backend/app/api/v1/auth.py: Moved _record_attempt() from finally block into except HTTPException and except Exception handlers
- backend/tests/test_auth.py: Updated test_register_rate_limit_triggers_after_three_attempts to use duplicate-email failures instead of successful registrations
- backend/tests/conftest.py: Included TestClient.response fix to unblock backend tests

Impact it Made:
- Rate limiter now correctly only counts failed registration attempts
- Prevents successful registrations from diluting the rate limit counter
- All 17 auth tests pass locally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,4 +119,4 @@ pdf_export.txt
 rag_ingest.txt
 
 # Models
-backend/app/modules/guard/models/*
+backend/app/modules/guard/models/*backend/rag_documents/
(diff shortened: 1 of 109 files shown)


SdSarthak:
Duplicate of #1174 — same fix to move registration rate-limit record out of `finally` block. Please close this and rebase #1174 on `main`.

PR #1326, opened 2026-07-01
fix : record registration rate-limit attempt only on failure

### LINKED ISSUE #1306: fix : record registration rate-limit attempt only on failure, not in finally block
In POST /api/v1/auth/register (backend/app/api/v1/auth.py), the rate-limit attempt is recorded in a finally block, meaning successful registrations also consume the rate-limit budget. After 3 successful registrations from one IP within the window, the IP is blocked for an hour - even though no actual abuse occurred.

Changes that Need to be Made:
- backend/app/api/v1/auth.py: Move the _record_attempt() call from the finally block into the except block so only failed registration attempts count toward the rate limit.

Impact that it would Provide:
- Successful registrations no longer consume rate-limit budget
- Consistent with login endpoint which correctly only records on failure
- Prevents legitimate users from being blocked after successful signups

tmdeveloper007:
Closes #1306.

Moved the _record_attempt() call from the finally block into the except
block in POST /register handler so only failed registrations count toward
the rate limit. This matches the login endpoint behavior and prevents
legitimate users from being blocked after successful signups.

Changes Made:
- backend/app/api/v1/auth.py: moved _record_attempt() from finally block
  into the generic Exception handler so it fires only on actual failures

Impact it Made:
- Successful registrations no longer consume rate-limit budget
- Consistent behavior with login endpoint which already records only on failure
- Prevents legitimate users being blocked after successful registrations

### DIFF
--- a/backend/app/api/v1/auth.py
+++ b/backend/app/api/v1/auth.py
@@ -184,6 +184,11 @@ def register(
         raise
     except Exception:
         db.rollback()
+        _record_attempt(
+            _auth_registration_attempts_by_ip,
+            client_ip,
+            _AUTH_REGISTER_RATE_LIMIT_WINDOW_SECONDS,
+        )
         # Generic database error handler
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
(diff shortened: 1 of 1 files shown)


SdSarthak:
Duplicate of #1174 — same fix to record rate-limit only on failure.

PR #1327, opened 2026-07-01
fix : commit GuardScanLog only after create_notification succeeds

### LINKED ISSUE #1304: fix : commit GuardScanLog only after create_notification succeeds in guard scan handler
In log_scan() (backend/app/api/v1/guard.py), when create_notification() fails and db.rollback() is called, the code attempts to re-add the log object with db.add(log). However, log was already committed at an earlier step, so after rollback it is in a detached state. Calling db.add() on a detached object causes IntegrityError on the next db.commit(), which is silently caught by the outer except block - losing the entire scan log entry silently.

Changes that Need to be Made:
- backend/app/api/v1/guard.py: Defer db.commit() for GuardScanLog until after create_notification() succeeds. If notification creation fails, rollback without trying to re-add the already-detached log.

Impact that it would Provide:
- Guard scan logs are never silently lost
- Consistent audit trail for compliance requirements
- IntegrityError becomes visible and actionable

tmdeveloper007:
Closes #1304.

Changed log_scan() to use db.flush() instead of db.commit() before
calling create_notification(). If notification creation fails, the
rollback correctly undoes the uncommitted log entry instead of trying
to re-add an already-committed detached object.

Changes Made:
- backend/app/api/v1/guard.py: replaced db.commit() + db.refresh() with
  db.flush() + db.refresh() so the log stays uncommitted until after
  notification creation succeeds

Impact it Made:
- Guard scan logs are never silently lost on notification failures
- No IntegrityError from re-adding a detached committed object
- Consistent audit trail for compliance requirements

### DIFF
--- a/backend/app/api/v1/guard.py
+++ b/backend/app/api/v1/guard.py
@@ -171,7 +171,7 @@ def log_scan(user_id: int, prompt: str, result: dict, ip_address: str | None = N
         log = _build_guard_scan_log(user_id, prompt, result, ip_address=ip_address)
 
         db.add(log)
-        db.commit()
+        db.flush()  # assign ID without committing
         db.refresh(log)
 
         if log.decision == "block":
@@ -189,8 +189,7 @@ def log_scan(user_id: int, prompt: str, result: dict, ip_address: str | None = N
             except Exception:
                 db.rollback()
                 logger.warning("Failed to create block notification for scan %d", log.id)
-                db.add(log)
-                db.commit()
+                raise
 
     except Exception:
         db.rollback()


SdSarthak:
This PR conflicts with recently-merged changes to `backend/tests/conftest.py` (via #1165). Please rebase on `main` and re-push — the changes here look good and will be reviewed for merge once the conflict is resolved.

tmdeveloper007:
Closing this PR. A cleaner version with the same changes (rebased onto current main) will be opened from a fresh branch.

